PR #51, opened 2025-04-05
Replace "solarflow_800_hyper" with "SF800" in displayed sensor name

SleXx88:
Improves readability in the Home Assistant Frontend: "solarflow_800_hyper" is only replaced by "SF800" for SF800 devices without influence.


FireSon:
Hi thanks for the update, but the actaul sensor name depends on the name you give the device in the Zendure App. It is not possible to do it in this way, since that would lead to duplicate names if you would have multiple SF800...

SleXx88:
That's right, you're right. I didn't pay attention to that.

PR #68, opened 2025-04-11
Update battery support +Fix

GeneralRurra:
# 🛠️ Feature Proposal: Improved Battery Sensor Handling for Zendure Integration

This update is based on recent improvements and thorough testing.  
It introduces multiple enhancements to improve sensor naming, value scaling, and error safety — especially for dynamically created battery sensors.

---

## 🚀 Summary of Improvements

- `packData` is received and processed via MQTT.
- Battery serial numbers (`sn`) are mapped to clean, user-friendly identifiers like `battery_1_`, `battery_2_`, etc., based on detection order.
- Entity names are now consistent and UI-friendly — no more raw serial numbers in the entity IDs.

- Voltage fields (`minVol`, `maxVol`, `totalVol`) are scaled from raw values (e.g., `332`) to real-world units (e.g., `3.32 V`).
- Temperature values (`maxTemp`) are converted from raw format (e.g., `2841`) to Celsius.
- Integer-based values like `softVersion` and `socLevel` are intentionally not scaled.

- Fixed a logic issue in `sensorAdd()` where the check `if value:` was incorrectly skipping valid values like `0`.  
  This has been corrected to:
  ```python
  if value is not None:
  ```

- Added a safeguard in `update_value()` to prevent the error:
  ```
  Error: Event loop is closed setting state...
  ```
  Now uses:
  ```python
  if self.hass and self.hass.loop.is_running():
      self.schedule_update_ha_state()
  ```

---

## 💬 Final Note

This update is based on personal improvements and real-world testing.  
I hope you’ll consider reviewing and including this in the project.

Thanks again for maintaining this great integration! 🙏

### DIFF
--- a/custom_components/zendure_ha/zenduredevice.py
+++ b/custom_components/zendure_ha/zenduredevice.py
@@ -22,6 +22,7 @@ from .number import ZendureNumber
 from .select import ZendureSelect
 from .sensor import ZendureSensor
 from .switch import ZendureSwitch
+from stringcase import snakecase
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -57,6 +58,9 @@ class ZendureDevice:
         self.topic_function = f"iot/{self.prodkey}/{self.hid}/function/invoke"
         self.mqtt: mqtt_client.Client
         self.entities: dict[str, Any] = {}
+        self.battery_sn_map: dict[str, int] = {}
+        self.battery_sn_counter = 1
+
         self.devices.append(self)
 
         self.lastUpdate = datetime.now()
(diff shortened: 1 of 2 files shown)


FireSon:
Nice thanks! 
Just checking, but some other people wanted to have 'seperate' HA devices for batteries, since they are more or less independant entities, and with entity showing the battery name. What do you think about that?

I will probably add your code manually, since I it has changed quite a bit and I want it in another file. But thanks a lot!!

SleXx88:
The battery is a very good idea. I also think that real values like “volts” etc. are easier to handle than pipe values.

Topic Energy Dashboard:
I changed the integration and ran on my test system. I added other sensors that calculate the power dissipation of the inverter. This already works. The sensor "PV Total" gives. Because the “PV Total device” sensor gives me a value without loss. This would falsify the statistics of the dashboard. The first sensors are already installed in the first tile.
Screenshot_20250411-175249

FireSon:
@GeneralRurra thanks for the input! I have added your code with a few small modifications in version 30

PR #88, opened 2025-04-13
Fix detection of maxVol, totalVol and minVol

Andi252:
The change in 2e0ce1c will not detect maxVol, minVol and totalVol as key.


FireSon:
Thanks for the PR!! In order to fix the battery entities properly I have moved the creation of the entities to another part of the program. I did add your change, but due to this change I cannot merge it. But it is much appreciated!

PR #121, opened 2025-04-20
Update energy sensors

harrymayr:
Always update both senors, to have no jumps in data and have a a reset at midnight (charging is not very likely then ;-)  )


GeneralRurra:
Even if it hasn't been published yet, the changes look good to me

image

image

harrymayr:
I wonder, that your aggrChargeDaykWh is negative

GeneralRurra:
Nene is not negative that is true of the Energy Dashboard

FireSon:
Thanks, seems logical, allthough I would think that I would allways get a value 0 anyway. Will merge it and change it a little bit.

PR #130, opened 2025-04-23
Hub 1200 & 2000 do not have AC charging mode.

danez:
I removed the AC switch from Hub1200 and Hub2000 as these devices are only connected to DC. This also means that the sensor for `gridInputPower` is useless.

### DIFF
--- a/custom_components/zendure_ha/devices/solarflow2400ac.py
+++ b/custom_components/zendure_ha/devices/solarflow2400ac.py
@@ -9,6 +9,7 @@ from homeassistant.core import HomeAssistant
 
 from custom_components.zendure_ha.binary_sensor import ZendureBinarySensor
 from custom_components.zendure_ha.number import ZendureNumber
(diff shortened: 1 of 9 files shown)


custom_components/zendure_ha/devices/solarflow2400ac.py
```diff
@@ -18,7 +19,7 @@
 
 class SolarFlow2400AC(ZendureDevice):
     def __init__(self, hass: HomeAssistant, h_id: str, definition: ZendureDeviceDefinition) -> None:
-        """Initialise AIO2400."""
+        """Initialise SolarFlow2400AC."""
         super().__init__(hass, h_id, definition, "SolarFlow 2400 AC")
         self.powerMin = -2400
```
danez, line 24: What is `powerMin`? The Hub2000&1200 have it set to -1000, which I'm not sure what that means. Is that supposed to be the AC charging power?

PR #134, opened 2025-04-23
Add support for Ace 1500

danez:
The Ace can be a subdevice of the Hubs. I'm not 100% sure if this works I do not have an Ace. I just looked at other projects :) And also checked the HTTP requests from the App.

### DIFF
--- a/custom_components/zendure_ha/api.py
+++ b/custom_components/zendure_ha/api.py
@@ -109,6 +109,28 @@ class Api:
         _LOGGER.error(response.text)
         return None
 
+    async def _createDeviceDefinition(self, data: dict[str, str], parent: ZendureDeviceDefinition | None = None) -> ZendureDeviceDefinition | None:
+        prodName = data.get("productName")
+        if not data or not (deviceKey := data.get("deviceKey")) or not prodName:
+            _LOGGER.debug(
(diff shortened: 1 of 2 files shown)


FireSon:
Nice 😄 
Two questions:
1) What use is there if we make the connection between these devices? (Except to know they are connected) is there additional functionality? Do they work together?
2) You create a subdevice, are you sure that the ACE1500 is not also in the list itself? If so it would suffice to add a parentId to the ZendureDeviceDefinition. Should the actual connection not be made between the actual devices? The ZendureDeviceDefinition is used to persist the device information, it is not the device itself.

kutzera:
It would be great if ACE 1500 could be implemented as well.
If you connect the ACE and HUB the values for remaining charge and discharge time and die gridinput power are not transmitted via the HUB. these values only come through the ACE entities.

FireSon:
@danez Again many thanks for your PR!! Since a lot has changed, especially with the addition of the batteries, I took from it what I thought was the essence and added it to the develop branch. I hope you check back in to see it I did it correctly.

@CWILDMOUNTAIN @kutzera Hopefully you can test the 41-pre2 version when it is out, to see if at least we have the correct entities. Not sure of the specific capabilities when used together with the Hub2000. But if the entities are there we can look for additional functionality.

PR #137, opened 2025-04-23
Update energy sensors & Limit setpoint on small PV values

harrymayr:
Update energy sensors:
Values with 0 need to be taken into account, otherwise a long time between two non zero numbers create a wrong wattHour, if the power goes to 0 inbetween for this long time (totalValue will kept on the last non zero value).
And reset to 0 at midnight also must be done, if the actual value is 0.

Limit setpoint on small PV values:
The goal is to set p1 to -SmartMode.MIN_POWER. 
But if you add SmartMode.MIN_POWER in case of actual power + p1 is smaller than SmartMode.MIN_POWER, it will switch to Unload. So the setpoint should not be too large.


harrymayr:
It now resets aggrChargeDaykWh to 0. Maybe a little bit later, because it will not updated that often, if the value is zero (every ~90sec?)
Zendure HA

harrymayr:
And with respect to limit the setpoint, the changes show now switching back and force anymore (SmartMode.MIN_POWER set to 30)
Zendure HA2

custom_components/zendure_ha/zendurermanager.py
```diff
@@ -329,7 +329,7 @@ def _update_smart_energyp1(self, event: Event[EventStateChangedData]) -> None:
             if self.operation == SmartMode.MANUAL:
                 self.updateSetpoint(self.setpoint, ManagerState.DISCHARGING if self.setpoint >= 0 else ManagerState.CHARGING)
             elif powerActual < 0:
-                self.updateSetpoint(min(0, powerActual + p1 + SmartMode.MIN_POWER), ManagerState.CHARGING)
+                self.updateSetpoint(min(0, powerActual + p1 + min(abs(powerActual),SmartMode.MIN_POWER)), ManagerState.CHARGING)
```
FireSon, line 332: I do not think this is necessary, since the updateSetpoint routine applies a max(0, power) to the power. So it would never discharge even if this value > 0

harrymayr: Without the limitation I had the behaviour showed in this [post ](https://github.com/FireSon/Zendure-HA/pull/115#issuecomment-2816781763)
It demand more power to charge as it is available at that time, so p1 will get positive in one of the next checks, will switch to Unload or Idle and than again to charge. With the limitation I didn't see that behaviour anymore (and it does not hurt ;-)  )

FireSon: I understand, but maybe I am a bit of a purist that I do not add things if I do not understand it 😄 The reason to add MIN_POWER is that the inverters themself use 50w so it does not make sense to charge below 50w. But you are right that at the moment once the inverter starts, it immediately shuts down if the power drops 1 watt, which is not good! I have added:
    powerActual = powerActual + p1 + (SmartMode.START_POWER if powerActual == 0 else SmartMode.MIN_POWER)
    self.updateSetpoint(min(0, powerActual), ManagerState.CHARGING)
That should result in a buffer of 50w before the inverter shuts down again. Would that solve the problem?

FireSon: Well the 50w is something Zendure does as well if you use their CT Smart mode. It seemed logical to follow their way of charging. But you are right that in this way 50w are 'lost' to the grid. But you could regard that as a 'buffer' which is used to prevent consumption from the grid if there is a drop in solar input. The power you consume from the grid is much more costly right?
The ManagerState.IDLE is only used to switch between charging and discharging. There is a delay of 10 seconds before it actually switches. In the past the smart mode was ended, which resulted in some unexpected bahaviour. Now the smart mode remains active on the device. I have done quite some test with this mode, but it reacts very quickly! The update interval for a new setpoint is 3 seconds for large changes (>100w), and 5 seconds for smaller ones.

harrymayr: > ... But you are right that in this way 50w are 'lost' to the grid. But you could regard that as a 'buffer' which is used to prevent consumption from the grid if there is a drop in solar input. The power you consume from the grid is much more costly right? ...

 And for cluster devices 50W each. Maybe let the user choose this value and make it a changeable setting with a slider or so.
With the argumentation of preventing consumption from the grid, you need to add also an offset for discharging 😄
I'm still not convinced, if your solution is the better one. I agree to add a buffer, but I don't see the need for the START_POWER.

But for sure, you need to allow to reset the energy sensors even if the value (totalValue) is zero. Otherwise the aggrChargeDaykWh sensor will be reset late (when charging will start in the morning).

harrymayr: I don't think the current version works the way I'd like either. It cycles the charging because there's not enough PV power available.
Zendure HA 8
Let's create a model, decide which behavior is best, and implement it. ;-)
Feed in

To be clear: This isn't about the offset itself (perhaps the value ;-)), but about the behavior at low PV power.

There are at least this 4 posibilities, if the available PV-power rises slowly, or keep in the range of 50-100W.
I like most the upper left or lower right.

What do you think?

FireSon: Well I want to be pragmatic. I have not studied it like you did 😄 
For me a few things are important, and let's choose the one who best fits these criteria:
1) Switching between discharging/discharging due to changing solar capacities should be limited. 
2) Once the communication is stable with the hyper2000, updating each second should be possible
3) Since buying electricity is more expencive than solar electricity, buying electricity should be avoided as much as possible.
4) Using the inverter will also cunsume energy, which should be accounted. Once it is started and there is still a surplus it should not matter, since it is in the p1 meter value

Did I forgot anything? Which of the models would fit the best...

harrymayr: I would rate point 3 pretty low, because if updating is fast enough, the amount of buying electricity is rather low.
Wrt point 1 I agree, question is, if switching happens already, if the hyper is set to idle (from charging). 
So perhaps the first one is the best, I will go for a test with this code:

`if self.operation == SmartMode.MANUAL:`
`    self.updateSetpoint(self.setpoint, ManagerState.DISCHARGING if self.setpoint >= 0 else ManagerState.CHARGING)`
`elif powerActual + p1 < SmartMode.MIN_POWER:`
`    self.updateSetpoint(min(0, powerActual + p1 + SmartMode.MIN_POWER), ManagerState.CHARGING)`
`elif powerActual > 0:`
`    self.updateSetpoint(max(0, powerActual + p1), ManagerState.DISCHARGING)`
`elif self.zero_idle == datetime.max:`
`    _LOGGER.info(f"Wait 10 sec for state change p1: {p1}")`
`    self.zero_idle = time + timedelta(seconds=SmartMode.TIMEIDLE)`
`elif self.zero_idle < time:`
`    if p1 < SmartMode.MIN_POWER:`
`        self.updateSetpoint(p1, ManagerState.CHARGING)`
`        _LOGGER.info(f"Update state: p1: {p1}")`
`    elif p1 >= 0`
`        self.updateSetpoint(p1, ManagerState.DISCHARGING)`
`        _LOGGER.info(f"Update state: p1: {p1}")`
`    else`
`        _LOGGER.info(f"Wait 10 sec for state change p1: {p1}")`               
`    self.zero_idle = datetime.max`

PR #139, opened 2025-04-24
Add support for changing bypass mode on Hub 1200/2000

danez:
I tested this with Hub2000, and it works. I assume that Hub1200 and Hyper2000 work the same, but I'm not sure about the newer devices.

Also, out of curiosity, the code mentions a Hub1000. What is that? I have never heard of this device.

### DIFF
--- a/custom_components/zendure_ha/devices/hyper2000.py
+++ b/custom_components/zendure_ha/devices/hyper2000.py
@@ -36,6 +36,7 @@ class Hyper2000(ZendureDevice):
             self.binary("wifiState", None, "switch"),
             self.binary("heatState", None, "switch"),
             self.binary("reverseState", None, "switch"),
+            self.binary("pass", None, "switch"),
             self.binary("lowTemperature", None, "switch"),
         ]
         ZendureBinarySensor.addBinarySensors(binaries)
@@ -72,7 +73,6 @@ class Hyper2000(ZendureDevice):
             self.sensor("gridInputPower", None, "W", "power", "measurement"),
             self.sensor("packInputPowerCycle", None, "W", "power", "measurement"),
             self.sensor("outputHomePowerCycle", None, "W", "power", "measurement"),
-            self.sensor("pass", None),
             self.sensor("socStatus", None),
             self.sensor("strength", None),
             self.sensor("hyperTmp", "{{ (value | float/10 - 273.15) | round(2) }}", "°C", "temperature", "measurement"),
--- a/custom_components/zendure_ha/devices/hub1200.py
+++ b/custom_components/zendure_ha/devices/hub1200.py
@@ -9,6 +9,7 @@ from homeassistant.core import HomeAssistant
 
 from custom_components.zendure_ha.binary_sensor import ZendureBinarySensor
 from custom_components.zendure_ha.number import ZendureNumber
+from custom_components.zendure_ha.select import ZendureSelect
 from custom_components.zendure_ha.sensor import ZendureSensor
 from custom_components.zendure_ha.zenduredevice import ZendureDevice, ZendureDeviceDefinition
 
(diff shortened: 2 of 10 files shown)


Wunni73:
@danez  "pass_mode": {
         "name": "Bypass-Modus",
         "state": {
           "auto": "Automatisch",    # = 0 Auto 
           "off": "Immer aus",          # = 1  Manual 
           "on": "Immer an".             # = 2 An (ON)

das mit den immer an , finde ich die Bezeichnung schlecht, weil wenn die Batterie voll ist geht er in Bypass  2 = Bypass An ( ist aber immer noch in Automatik )

danez:
Soweit ich das verstanden habe ist `passMode` nur der Status der Einstellung, also genau wie in der App.
Wenn der Wechselrichter in den Bypass geht, dann ändert sich `pass` von 0 auf 1.

Hast du das getestet? Also dass sich `passMode` ändert wenn der Wechselrichter automatisch in Bypass geht?

Wunni73:
:-) der geht täglich um 15 Uhr in 2 ansonsten ist er in 0
Bildschirmfoto 2025-04-24 um 20 09 42

reverseState ist das gleiche nur 0 off und 1 on, auch der Bypass

danez:
> Wunni73

Okay ich werd am Wochenende testen.

danez:
Auf deinem Screenshot ist aber `pass` zu sehen, nicht `passMode`. Aber trotzdem interessant das `pass` auch 2 sein kann.

custom_components/zendure_ha/devices/hyper2000.py
```diff
@@ -84,6 +84,10 @@ def sensorsCreate(self) -> None:
                 "acMode",
                 {1: "input", 2: "output"},
                 self.update_ac_mode,
+            ),
+            self.select(
```
FireSon, line 88: When you use a select, it means that the value can be manipulated is that the case? Or is this the resulting status? So either we must be able to change the value with the select, or use a sensor with the correct state names

Wunni73:
ich kenne nur Pass , PassMode habe ich beim Hyper 2000 noch nicht gesehen.

custom_components/zendure_ha/devices/hyper2000.py
danez: Yes `passMode` is the setting for the bypass and `pass` is showing the current status of the bypass.

FireSon:
@danez I think @Wunni73 is right, the hyper2000 does not have passMode

danez:
> @danez I think @Wunni73 is right, the hyper2000 does not have passMode

Okay, also found [this](https://forum.iobroker.net/topic/69579/test-adapter-zendure-solarflow/826?_=1745529281843&lang=en-US), where people report the same problem with iobroker.

 I removed the select for the hyper 2000 again.

custom_components/zendure_ha/devices/hyper2000.py
FireSon: Yes I understand, but when a user selects another value, you probably should update a property in the device. For the ac mode the routine self.update_ac_mode actually changes the ac mode

danez: This is happening directly in `ZendureDevice.select` method. I added the code there that if no custom `onwrite` is supplied it simply sets the property in mqtt.

https://github.com/FireSon/Zendure-HA/pull/139/files#diff-512fd90e9e7885a2c6fb1ca77f8d395d581dee2fde051adc54f7527e77a4dc58R278

FireSon: Ahh ok, I missed that, seems like a clever solution!

z-master42:
This shows very clearly that not all Zendure values can be lumped together, as they have been given different values over the various device generations.

PR #165, opened 2025-04-27
README update

Wunni73:
README config window

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,9 @@ After you have successfully installed and configured HACS you can simply press t
 
 ## Configuration options
 
-![image]
+<img width="494" alt="Bildschirmfoto 2025-04-27 um 10 11 14" src="https://github.com/user-attachments/assets/e805e120-56d6-47b7-abe1-236e80d4b2cd" />
+
+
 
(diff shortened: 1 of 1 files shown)


z-master42:
What is Hub 1000?

And while you're at it (and I don't have to fork the repo):

In the description of the exposed sensors (which is probably not up-to-date either), the descriptions of the Pack Input Power and Output Pack Power sensors are reversed. Power flows are always described from the perspective of the main device (hub/hyper).

Input is therefore power from the battery into the hub/hyper and output is the power into the battery.

FireSon:
Thanks for the input, a few questions/remarks:

1) To be honest I do not know what the Hub1000 is, somebody asked me to add it. It it is not an existing device I will take it out.
2) The update of the configuration screenshot is great!
3) Why would you remove the information about the cluster configuration? When you have more tahn one device it is important to explain.

I think it is wise to put more info into a wiki. Also the device specific properties. Use the main readme as a way to find the necessary information...

FireSon:
Since I will refactor the Readme soon, and add new screenshots, I will close this for now

PR #176, opened 2025-04-30
Avoid switching on low PV power

harrymayr:
For me, it looks now good. Of course at the moment I don't have a small rising of the PV-power ;-) but on Sunday it may show how it behave with smaller gradients.
I think we can close the other commit.

image

### DIFF
--- a/custom_components/zendure_ha/zendurermanager.py
+++ b/custom_components/zendure_ha/zendurermanager.py
@@ -340,17 +340,23 @@ class ZendureManager(DataUpdateCoordinator[int]):
             # update the setpoint
             if self.operation == SmartMode.MANUAL:
                 self.updateSetpoint(self.setpoint, ManagerState.DISCHARGING if self.setpoint >= 0 else ManagerState.CHARGING)
-            elif powerActual < 0:
-                powerActual = powerActual + p1 + (SmartMode.START_POWER if powerActual == 0 else SmartMode.MIN_POWER)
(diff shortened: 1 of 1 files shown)


FireSon:
I have looked at the power distribution once more, since a lot of people have reported problems with big power changes. I think I found one reason for it and fixed it. Unfortunately that changed some of your code as well. Can you have a look at it? Basically now it first stops charging/discharging and once in idle it will start depending on the p1 value.
In order to make it more readable I added some comments as well

FireSon:
I will do my updates via pull request as well, just lazyness I imagine 😄

harrymayr:
OK, updated to 1.0.41pre this afternoon. I will check tomorrow morning. Now, you have no buffer anymore on charging and it will regulate to 0W which may get a larger amount of switching. I would change this lines:

` # update when we are charging`
`  elif powerActual < 0:`
`     self.updateSetpoint(min(0, powerActual + p1), ManagerState.CHARGING)`

to that what I proposed:

` # update when we are charging`
`  elif powerActual + p1 < SmartMode.MIN_POWER:`
`     self.updateSetpoint(min(0, powerActual + p1 + SmartMode.MIN_POWER), ManagerState.CHARGING)`

But let's see how it look like tomorrow. On fast rising of the solarpower it may not be a problem, but if it takes several minutes to get from i.e. 300W to 350W, there is a possibility of higher amount of switching.

PR #177, opened 2025-04-30
Update power if difference to previous update is > 1

harrymayr:
Over the day, I see many updates which only differs by +/-1, especially if the charging or discharging is on it's maximum. This updates are not necessary and can reduce communication load.

image


FireSon:
Is the check:
`if delta > 1 and inprogram:`
not the other way around? Should it not be:
`if delta <= 1 and inprogram:`

harrymayr:
> Is the check: `if delta > 1 and inprogram:` not the other way around? Should it not be: `if delta <= 1 and inprogram:`

Yes, the first commit is wrong, I changed it in the later commit.

PR #179, opened 2025-05-01
Update en.json

twoenter:
correction sensor names:
Temperatures are current temperatures, not maximum 
out and input sensornames
Hyper temperature not battery temperature

### DIFF
--- a/custom_components/zendure_ha/translations/en.json
+++ b/custom_components/zendure_ha/translations/en.json
@@ -249,7 +249,7 @@
         "name": "Battery 4 Current"
       },
       "pack_input_power": {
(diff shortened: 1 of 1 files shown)


custom_components/zendure_ha/translations/en.json
```diff
@@ -249,7 +249,7 @@
         "name": "Battery 4 Current"
       },
       "pack_input_power": {
-        "name": "Battery Output"
+        "name": "Battery Input"
```
harrymayr, line 252: That's not correct. pack_input_power is the power which comes from the battery into grid. So it is Battery Output.

PR #182, opened 2025-05-02
Calculate device total capacity

harrymayr:
Assuming, your solution will look similar ;-)
For now, it's from my side just a guess, that AB1000 serial starts with A.. and Ab2000 with C..


FireSon:
I just pushed a lot of changes, including the capacity. Due to the many changes I cannot merge this PR
But many thanks, as allways 😄

PR #194, opened 2025-05-04
Hub 1200/2000 first cleanup + SoH added to batteries

z-master42:
Added SoH to the batteries and removed definitely non-existent sensors from the Hub 1200 and 2000.

In this context, I also don't know what reverseState is and I've never heard of energyPower.

### DIFF
--- a/custom_components/zendure_ha/zendurebattery.py
+++ b/custom_components/zendure_ha/zendurebattery.py
@@ -32,6 +32,7 @@ class ZendureBattery(ZendureBase):
             self.sensor("state"),
             self.sensor("power", None, "W", "power", "measurement"),
             self.sensor("socLevel", None, "%", "battery", "measurement"),
+            self.sensor("soh", "{{ (value / 10) }}", "%", None),
             self.sensor("maxTemp", "{{ (value | float/10 - 273.15) | round(2) }}", "°C", "temperature", "measurement"),
             self.sensor("softVersion"),
         ]
--- a/custom_components/zendure_ha/devices/hub1200.py
+++ b/custom_components/zendure_ha/devices/hub1200.py
@@ -49,20 +49,17 @@ class Hub1200(ZendureDevice):
         sensors = [
             self.sensor("hubState"),
             self.sensor("solarInputPower", None, "W", "power", "measurement"),
-            self.sensor("BatVolt", None, "V", "voltage", "measurement"),
             self.sensor("packInputPower", None, "W", "power", "measurement"),
             self.sensor("outputPackPower", None, "W", "power", "measurement"),
             self.sensor("outputHomePower", None, "W", "power", "measurement"),
             self.sensor("remainOutTime", "{{ (value / 60) }}", "h", "duration"),
             self.sensor("remainInputTime", "{{ (value / 60) }}", "h", "duration"),
-            self.sensor("soH", "{{ (value / 10) }}", "%", None, "measurement"),
             self.sensor("packNum", None),
             self.sensor("electricLevel", None, "%", "battery"),
             self.sensor("energyPower", None, "W"),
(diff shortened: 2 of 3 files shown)


custom_components/zendure_ha/zendurebattery.py
```diff
@@ -32,6 +32,7 @@ def entitiesCreate(self) -> None:
             self.sensor("state"),
             self.sensor("power", None, "W", "power", "measurement"),
             self.sensor("socLevel", None, "%", "battery", "measurement"),
+            self.sensor("soh", "{{ (value / 10) }}", "%", None),
```
harrymayr, line 35: At least the Hyper 2000 does not deliver a SoH value, so this is not the right way to do it.

z-master42: In the device...py it is also wrong, as it is a battery value and not a device value.

I can say the same for batcur, because this battery value is not available via the two hubs. So that's not the way either.

FireSon: @z-master42 Nice to see these improvements 😄 
I understand the line of thought regarding the SOH, it is however reported by the Hub, and probably if you have more than one battery attached, calculated for the Hub as a whole. Therefore the right place would be in the hub1200.py and the hyb2000.py I would say that if Zendure reports it as a battery property, it is part of the ZendureBattery, otherwise it is a device property. 
Hope that helps!

But your input is much appreciated!

z-master42: I can't confirm that. The SoH is displayed separately for each battery and there is no summary at hub level.

FireSon: OK then it should be part of the ZendureBattery. Since I know that the Hyper2000 does not report this property, I must find a way to make sure it does not show up there because otherwise we will get a lot of questions! 
Let me think about is, will merge your PR for now

z-master42: Thank you. It's exactly the same the other way round with batcur. In HA you can deactivate the entities if necessary 😉.

I have an idea in my head, but I'll have to see for myself later whether it can be implemented.

FireSon:
I have fixed the initialization, by adding a seperate method per device to add device specific entities to a battery.

z-master42:
```
def entitiesBattery(self, sensors: list[ZendureSensor]) -> None:
        sensors.append(self.sensor("soh", "{{ (value / 10) }}", "%", None))
```
That does not work. This only creates a soh entity in the hub device, the status of which is `unknown`.

The soh entity in the battery devices is of course present, but is now unformatted again.

FireSon:
Yes I saw it, fixed it in the mean time

z-master42:
Ah, ok. I must have overlooked that. I have set another PR in the meantime.

harrymayr:
I see some difficulties to add only the values (i.e. Current, SoH) for devices we know that they show them. Do you know for every device, if they are available or not? Perhaps you can open a Wiki and/or discussions page in github, were we can put the information we know already. i.e. which values are in principle available (for some of the devices). This information can be filled with the information we know and mark the one, we don't know for now. But for values/sensors we don't know if they are available, we need to show them anyway, even if the show 'unknown'. If then someone give the information, we can change it.

FireSon:
Yeah I think the wiki is defenitly something helpful the readme is to large and people do not read it 😉

z-master42:
If I have understood this correctly, the integration returns all entities that appear on the Zendure side via MQTT, regardless of whether they have been further specified in a ...py or not. This means that everything has a value, even if it is only a 0. In addition, there are many entities that are not visualised or available as a function on the app's user interface.

So if an entity has been specified in a ...py, but permanently displays unknown, it will not be delivered for this device on the Zendure side.

I also think that a wiki is a good idea here.

PR #197, opened 2025-05-05
Sensors removed or reconfigured

z-master42:
- SoH entity added to a battery if the parent device is a Hub 1200 or Hub 2000.
- Show `batcur` only if the parent device is a Hyper 2000. No idea whether this also applies to the newer devices.
- `masterSwitch` and `buzzerSwitch` changed from binary to switch for the Hubs, Hyper and AIO. `buzzerSwitch` toggles the confirmation tone on and off.
- `autoRecover` changed from binary to switch for the Hubs. This can be used to specify whether the bypass mode should switch back to its default setting (Automatic) the next day.
- `inverseMaxPower` changed from sensor to number for the hubs. This is the value that defines the maximum power of the connected inverter.
- `reverseState` and `energyPower` removed for the hubs, as the value here is permanently unknown.

The localization JSON still needs to be adapted.

### DIFF
--- a/custom_components/zendure_ha/devices/aio2400.py
+++ b/custom_components/zendure_ha/devices/aio2400.py
@@ -44,6 +44,8 @@ class AIO2400(ZendureDevice):
 
         switches = [
             self.switch("lampSwitch", None, "switch"),
+            self.switch("masterSwitch", None, "switch"),
+            self.switch("buzzerSwitch", None, "switch"),
         ]
         ZendureSwitch.addSwitches(switches)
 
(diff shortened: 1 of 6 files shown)


FireSon:
Well sorry but I did do a lot of changes! Also the way to add device specific entities I was working on has my preference. 

You made a few sensors editable, basically I am fine with that but would however like to know what they do.
Like the MasterSwitch and buzzerSwitch. 

The InvertMaxPower was not editable by choice, since I do not want to make it very easy to bypass the 'legal' responsibilities. If however there are compelling reasons to be able to edit them please let me know 😄

z-master42:
All good.

I don't know what masterSwitch is doing, but it only shows switched on as binary.

buzzerSwitch is available to switch the beep tones of the device on and off.

passMode defines the bypass mode of the hub and recoverState is included here. As I have already described initially.

All functions that are also available in the app.

PR #200, opened 2025-05-05
Fix description of AB2000(S) battery

harrymayr:


FireSon:
Mmm I already fixed that.. but apparently not saved...

PR #204, opened 2025-05-06
Update hyper2000 charging capabilitys

harrymayr:
image

### DIFF
--- a/custom_components/zendure_ha/devices/hyper2000.py
+++ b/custom_components/zendure_ha/devices/hyper2000.py
@@ -83,7 +83,6 @@ class Hyper2000(ZendureDevice):
         ZendureSelect.addSelects(selects)
 
     def entitiesBattery(self, battery: ZendureBase, sensors: list[ZendureSensor]) -> None:
-        sensors.append(battery.sensor("soh", "{{ (value / 10) }}", "%", None))
         self.batCount += 1
         self.powerMin = (-1200 if battery.kwh == 2 else -950) if self.batCount == 1 else -1600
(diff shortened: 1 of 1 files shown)


FireSon:
@harrymayr I do not think this is correct. 
The screenshot you are reverring to has to do with solarpower. The max. input power for AC charging is 1200 watt. Can you check if that is correct? Also the 'calculation' of the kWh of the batteries, if it is necessary, can be much simpler since the ZendureBAttery has a kWh property.

harrymayr:
Unfortunatly you are right ;-)
I close this.

harrymayr:
But do we control the AC input / output or do we control the charging (also from PV). In this case we will limit the charging even with solar to 1200W?

harrymayr:
I tested this commit with a power supply on the PV input and this allow charging with higher power than 1200W. 
Screenshot_20250508_071923_Zendure

FireSon:
We can only regulate the AC, but it is certainly something we must look into once I get my solarpanels going. It probably changes the amount of power you can sent to the device! Close this for now.

harrymayr:
I will check the Effekt of the PV input with your implementation tomorrow. My assumption is, that the maximum charging power is then 1200W, where 80W comes from the solar input and 1120W from the grid.
I assume, that we control the power in and out of the batterie. As I know, the HUB 1200 has no  AC input and we control the batterie (dis-)charge.

harrymayr:
You are right, with the pre6 it charges also more than 1200W if I switch on my power supply on the PV input. I wonder, if we need to limit AC power, if there is more than 400W PV input, or if the Hyper will do this on it's own.

PR #205, opened 2025-05-06
Update hub1200 charging capabilities

harrymayr:
The HUB 1200 can charge with 1200W if at least one AB2000 is connected.

### DIFF
--- a/custom_components/zendure_ha/devices/hub1200.py
+++ b/custom_components/zendure_ha/devices/hub1200.py
@@ -21,7 +21,7 @@ class Hub1200(ZendureDevice):
     def __init__(self, hass: HomeAssistant, deviceId: str, prodName: str, definition: Any) -> None:
         """Initialise Hub1200."""
         super().__init__(hass, deviceId, prodName, definition)
-        self.powerMin = -1000
+        self.powerMin = -800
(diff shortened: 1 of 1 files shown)


custom_components/zendure_ha/devices/hub1200.py
```diff
@@ -69,6 +69,9 @@ def entitiesCreate(self) -> None:
 
     def entitiesBattery(self, battery: ZendureBase, sensors: list[ZendureSensor]) -> None:
         sensors.append(battery.sensor("soh", "{{ (value / 10) }}", "%", None))
+        if battery.kwh == 2:
```
FireSon, line 72: Every time a battery is added this routine is called. The ZendureDevice has also a kWh property, I might me better to use that one. If you have two AB1000 it is also -1200 right?

harrymayr: > If you have two AB1000 it is also -1200 right?

To my knowledge not (don't why) but @Linus86529 figured out that only if an AB2000 is connected 1200W is possible.

FireSon: Ah OK, just checking well then it is all fine 😄

PR #209, opened 2025-05-06
Add aggr sensors which do not reset. 

harrymayr:
I added 3 more sensors, which collect all the charge/discharge/solar data. Somehow a meter like the statistics in the Zendure app. These sensors do not reset at midnight.
In general I don't think it's necessary to have daily reset sensors (HA can do this job), but want to have it downwards compatible.

An additional idea is, to allow a preset of these sensors, so you can set it to the total values from the App. But my understanding of the config_flow is too limited to get this done ;-)

### DIFF
--- a/custom_components/zendure_ha/sensor.py
+++ b/custom_components/zendure_ha/sensor.py
@@ -100,7 +100,6 @@ class ZendureRestoreSensor(ZendureSensor, RestoreEntity):
             secs = time.timestamp() - self.lastValueUpdate.timestamp()
             self._attr_native_value = float(self.state) + self.last_value * secs / 3600000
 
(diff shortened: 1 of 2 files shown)


FireSon:
I agree! better to remove the daily ones, since multiple people complained about the faulty reset.

harrymayr:
> since multiple people complained about the faulty reset.

I think this is solved, I see only problems, that the sensors has large jumps for me in the statistic, due to my frequent restart of HA, to test the Zendure HA ;-)

FireSon:
I cherrypicked this PR, since I refactored qite a bit. Decided to remove the daily ones since it is not necessary and I want to keep it as simple as possible. Will be in the next release.

PR #222, opened 2025-05-10
Update zenduredevice.py

harrymayr:
Total_increasing does not work with last_reset (in sensor.py)


FireSon:
The whole idea is to not reset these sensors right? Keep them always increasing. Then it should be total_increasing so the energy dashboard will take care of the reset...

harrymayr:
This is also fine, but then this line needs to be removed (beside the check on the day ;-)  )

image

FireSon:
Fixed it in a more generic way, you can now use both

PR #223, opened 2025-05-10
Update zendurebase.py

harrymayr:
aggr sensor_state will stay None until it is updated in sensor.aggregate ;-) In my debug (sensor is ZendureRestoreSensor) was always false

However, this may not be the right change to get the sensors not reset to 0 on a restart. So feel free to program a better solution

### DIFF
--- a/custom_components/zendure_ha/zendurebase.py
+++ b/custom_components/zendure_ha/zendurebase.py
@@ -226,7 +226,7 @@ class ZendureBase:
 
     def aggr(self, name: str, value: int) -> None:
         """Aggregate value to sensor."""
-        if (sensor := self.entities.get(name, None)) and sensor.state is not None and sensor is ZendureRestoreSensor:
+        if (sensor := self.entities.get(name, None)):
             try:
                 time = dt_util.now()
                 sensor.aggregate(time, value)


FireSon:
Fixed this i a more generic way, but many thanks for the help 😄

harrymayr:
Any idea to allow a preset? For example the total values you can get out of the App?

FireSon:
What do you mean by preset and values from the app?

harrymayr:
If you want to have the same aggr data as in the App, you need to enter the actual values. Perhaps there is a way in HA, but everytime I try to change a state of a statistical value, it resume with the old one.

Gesendet von Outlook für Android<https://aka.ms/AAb9ysg>
________________________________

FireSon:
Ahh I see, you would like to import the history? That would be nice 😄 I will have a look at that once version 41 is stable! 
Did you check the changes I merged? Did I forget anything?

harrymayr:
> Did you check the changes I merged? Did I forget anything?

Will do now, but I have some more PR :-D

harrymayr:
> Ahh I see, you would like to import the history?

Not realy import hte history, but set the aggr sensors to this values.

harrymayr:
> Did I forget anything

Left some comments on your merged commits.

harrymayr:
> Ahh I see, you would like to import the history? That would be nice 😄 I will have a look at that once version 41 is stable!

Found some [info](https://github.com/nograx/ioBroker.zendure-solarflow/issues/84#issuecomment-2746120744) for that

PR #229, opened 2025-05-10
Update zendurebattery.py

harrymayr:
Not really necessary to show 2 digits, because changes are in 1 degree resolution anyway.


FireSon:
Sorry this is not the right temp calculation. I fixed it another way

harrymayr:
Your new calculation does not make a difference to the old one ;-) :-D

PR #230, opened 2025-05-11
Allow change of the Bypass setting

harrymayr:
This commits allow to change the Bypass settings. For the Hyper I could succesfully test it, for the HUB 1200 if do have logs that confirm, that there is no gridReverse but the passMode, so I'm pretty sure, that this will work.
For the other devices I have no information if passMode or gridReverse is used.

The first commits b4bd600 and 7100977 include changes for an offset on discharging and the change wrt to the aggr sensors. You don't need to use these 2 commtis, but I have no idea how to exclude them from a PR. :-|

image

### DIFF
--- a/custom_components/zendure_ha/devices/hub1200.py
+++ b/custom_components/zendure_ha/devices/hub1200.py
@@ -64,7 +64,10 @@ class Hub1200(ZendureDevice):
         ]
         ZendureSensor.add(sensors)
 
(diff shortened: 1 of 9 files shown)


FireSon:
Once again thanks for the change! I cherrypicked your PR and added it slightly different. The update_bypass is not necessary, and actually the default implementation (entityWrite) is better since it uses the name of the entity which is different for Hub and Hyper

PR #236, opened 2025-05-11
Update zendurermanager.py

harrymayr:
Add offset per device at discharge. Useful for user, does not get money for feeding into grid. Need changes in zenduredevice.py

### DIFF
--- a/custom_components/zendure_ha/zendurermanager.py
+++ b/custom_components/zendure_ha/zendurermanager.py
@@ -429,6 +429,7 @@ class ZendureManager(DataUpdateCoordinator[int], ZendureBase):
                         clusterPower -= pwr
                     else:
                         pwr = 0
+                    pwr += d.offset
                 power -= pwr
 
                 # update the device


FireSon:
I did not merge the offset, since I did not know the details. Withou the others does this PR still make sense?

harrymayr:
This PR need #235 because the offset is defined in zenduredevice.py
The idea is similar to the offset you had for charging with the reason of not spending power to the grid (and inverter need 50W to work). The smart regulation will not regulate to 0 with this change. Perhaps you can read the last posts in #180

harrymayr:
With this offset, you can set the batterie output to a value, that it's less likely that you feed batterie power into grid.
image

MacdiverOne:
Stand heute (Standardregelung ohne Offset): ca. 300Wh ungewollte Einspeisung bislang. Auch die Hyper brauchen ihre Zeit, um z.B. Spitzen der Klimaanlage glattzuziehen. 

Insofern (und aus Prinzip den Netzbetreibern gegenüber) wäre ein konfigurierbarer Offsetwert schon wünschenswert. Klar kann man den immer wieder bei jedem Release selbst einbauen, aber das wäre ja wieder "Gefrickel" (oder ein Fork) - beides nicht erstrebenswert.

FireSon:
Well I have not looked into the details, since the local MQTT takes up most of my time. 

One of my goals is to decrease configuration properties as much as possible, since they are a 'source for errors' if people do not know what they are doing. And what would happen if you have 4 or 5 devices? In your proposal you would apply all offsets (both charging and discharging), wouldn't the margin then not become too big? So I am not convinced it is the best solution. Inorder to decrease the unnecessary 'buying' of electricity, I am happy to add a static offset. You can specify the best value for that offset.

Once the communication is stable, there can be a lot of 'improvements' done. One of the options might be a priority in which devices to use for a quick change. One might be to descrease the minimum reaction time, which is now 3 seconds but the hyper can react much quicker (<1 second). All of these ideas have an impact on the 'power distribution'. If

harrymayr:
OK, what about adding on offset slider to the manager, which is only usable if "Smart Mode" is active? So we only have one adjustment. The original idea was to use per device, because the HUB seems to have a larger diff between packInputPower and outputHomepower. But it is already OK to add the offset to the setpoint.

harrymayr:
> I am happy to add a static offset. You can specify the best value for that offset.

I think the best value may be device dependend. And also, if you get money for feeding in, or not. So a fixed offset is also not the best solution.

> One of the options might be a priority in which devices to use for a quick change. One might be to descrease the minimum reaction time, which is now 3 seconds but the hyper can react much quicker (<1 second).

Yes, I already thought about that. Perhaps take a maximum slewrate (device dependend) into account. I assume that only the HUB family is that slow. We may can get this slewrate automatically by checking the requested power (setpoint) and the first report of the packInputPower or outputPackPower.

FireSon:
Well I think you are overcomplicating this whole problem, and to be honest you have lost me a bit in the process. The calculation of the power distribution is the 'heart' of the integration and I am not willing to make changes whithout a clear reason, cause and solution. I am aware that improvements can be made, like I mentioned in my last comment, but I am a strong believer in the KISS principle (Keep it stupid simple) and I do not want to rush into tweaking of details as long as much works needs to be done the backbone of the application. 
That being said, your input is much valued, and I am very happy with your involvement. So can we agree that you test your solution for lets say, two weeks, see how it plays out in different scenario's, and then we decide together how we incorperate it into the integration? The solution must also work for people who have no clue how to use HA, automations, let alone batteries setting, and wonder at every entity what really means...

The original offset I had was for charging only, in your last PR it is only used for discharging. What would, in your opinion, be the best offset for charging and discharging at this moment? Than we can add it as a static value in the current version for now.

MacdiverOne:
From my point of view an offset of 20W for output to grid should do the trick for now and could be monitored for a while as baselining for further actions.

FireSon:
Does this offset mean that you want to discharge 20 watt 'extra' or less?

MacdiverOne:
Less than the "regular" calculated output would be if tried to get close to zero. For example: if 200W would get you close to zero import from grid, the offset will set 180W instead of 200W to reduce the risk of spending your valuables (specially in winter) for the sake of god (or your "Netzbetreiber").

harrymayr:
As always, you can argue in both directions:
In summer, where you don't get your fully charged batteries empty over night, it's usfull to spent more into grid, than receiving it.
In winter, where your battery will not fully charged, and you need to get some power from the grid anyway, you wont spent no single Watt of your batteries into the grid.

harrymayr:
> So can we agree that you test your solution for lets say, two weeks, see how it plays out in different scenario's, and then we decide together how we incorperate it into the integration?

@MacdiverOne(?) and @Linus86529 test this allready and are happy with this.
However, most user may not need or use this feature. I have it also applied in my installation, but have set it to 0.
Perhaps it would be an idea to have some "Advanced" setting, which could be enabled in configuration.yaml or through the config_flow.
I think the local mqtt is nice to have but not necessary for a bunch of users. This could be also an advanced setting which don't confuse "normal" 😉 users, if not visible.

FireSon:
> I think the local mqtt is nice to have but not necessary for a bunch of users. This could be also an advanced setting which don't confuse "normal" 😉 users, if not visible.

Well I do not really appriciate that, since it is not a totally fair comparison. Many people do not want to 'share' their data in the cloud. To be able to disconnect from the cloud is, in my opinion, a great selling argument to use HA. In regard to it being 'complex'; I have managed to enable/disable local MQTT with only one checkbox, making it available to also the less experienced HA users!

Optimizing the 'loss' of energy, depends on a lot of parameters, which makes it very tricky, and like you said, "you can argue in both directions". I do not want to go into that at this moment, and I have tried to explain why. Over the summer there is plenty of time to find the best way to deal with it.

PR #243, opened 2025-05-12
Update zendurebattery.py

harrymayr:
Format BMS firmware (softVersion)


FireSon:
@harrymayr thanks again 😄 
It is easier for me if you combine related changes in one PR instead of multiple, is that possible? But anyway much apreciated!

PR #275, opened 2025-05-18
Time update

harrymayr:
I added 2 sensors which show the charge/dischargeing time to min/max SOC. They also show "unkown" instead of 999 if the corresponding power = 0.
Added the translations for this sensors and some other.

### DIFF
--- a/custom_components/zendure_ha/translations/de.json
+++ b/custom_components/zendure_ha/translations/de.json
@@ -16,7 +16,7 @@
           "password": "Zendure Passwort",
           "p1meter": "P1 Sensor für Smart Matching",
           "mqttlog": "MQTT-Kommunikation loggen",
-          "mqttlocal": "Lokalen Mosquitto MQTT AddOn verwenden",
+          "mqttlocal": "Lokalen MQTT verwenden",
           "wifissid": "Wifi SSID",
           "wifipsw": "Wifi Passwort"
         }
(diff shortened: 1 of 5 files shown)


custom_components/zendure_ha/zenduredevice.py
```diff
@@ -119,6 +142,20 @@ def entityWrite(self, entity: Entity, value: Any) -> None:
 
         self.writeProperties({property_name: value})
 
+    def updateOutTime(self) -> None:
+        try:
+            if self.minSoc is not None and self.kwIn is not None and self.actSoc is not None :
+                self.setvalue("remainOutTime2minSoc", min(999,max(0,(float(self.kwh)*960 / max(10,float(self.kwIn)) *(float(self.actSoc)- float(self.minSoc))/100))) if float(self.kwIn) > 10 else None)
```
djodjo02130, line 148:
```suggestion
        # Early return if any required value is missing
        if not all([self.minSoc, self.kwIn, self.actSoc]):
            return
            
        power_input = float(self.kwIn)
        
        # Early return if power is too low
        if power_input <= 10:
            self.setvalue("remainOutTime2minSoc", None)
            return
        
        # Calculate discharge time to minimum SoC (in minutes)
        remaining_charge_pct = (float(self.actSoc) - float(self.minSoc)) / 100
        discharge_minutes = float(self.kwh) * 60 * remaining_charge_pct / power_input
        
        # Clamp to valid range
        result = min(999, max(0, discharge_minutes))
        self.setvalue("remainOutTime2minSoc", result)
```

djodjo02130: @harrymayr Hi ! Quick cleanup of `updateOutTime` method! 🧹

Made the code more readable by using early returns and breaking down the calculation into clear steps. Same functionality, just easier to follow.

Happy to discuss any feedback!

PR #292, opened 2025-05-24
Update zenduredevice.py

harrymayr:
BugFix aggr sensor names


FireSon:
Mmm thought I fixed that one 😄 thanks!

PR #293, opened 2025-05-24
Time update

harrymayr:
I changed things reflecting 1.0.41

### DIFF
--- a/custom_components/zendure_ha/devices/ace1500.py
+++ b/custom_components/zendure_ha/devices/ace1500.py
@@ -50,8 +50,6 @@ class ACE1500(ZendureDevice):
             self.sensor("packInputPower", None, "W", "power", "measurement"),
             self.sensor("outputPackPower", None, "W", "power", "measurement"),
             self.sensor("outputHomePower", None, "W", "power", "measurement"),
-            self.sensor("remainOutTime", "{{ (value / 60) }}", "h", "duration"),
-            self.sensor("remainInputTime", "{{ (value / 60) }}", "h", "duration"),
             self.sensor("packNum", None),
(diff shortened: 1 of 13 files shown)


FireSon:
@harrymayr Would it not be better to have only one remaining time? Do you think many people use remaining time? 
Personally I think we could just replace the existing remaining time...

harrymayr:
We also can remove the existing ones. With the change to the exact batterie capacities, the 960 in the calculation must be changed to 1000.

harrymayr:
Removed the old remaining times and changed the factor for calculating the time. But I have no idea how I can resolve the remaining conflicts :-|

harrymayr:
I think, this was one of my first approach. However, if you calculate with remainingOutTime, it will be shown in the sensor list? Then you have again 2*2 sensors for the remainig times. 
If you see a way to calculate with a sensor, but don't display him, it's of course easier (I know you like KISS :-D. )

FireSon:
Yes I like KISS very much😄 

Well I thought I made a PR, but it seems the changes are directly pushed to the main branch. But I basically replaced the value of the remaining times. It is comparable to an advanced template.  I renamed the ZendureVersionSensor to ZendureCalcSensor and created a calculation to adjust both remaining times. 

Can you check if the calculation of the remaining times are correct? Sorry to cherrypick your PR...

harrymayr:
> Sorry to cherrypick your PR...

No problem. My programming style is not always the best, I know that. It should work, not being beautiful :-D
I will check your changes.

harrymayr:
Slight differences:

image

harrymayr:
Also for a longer time, no real significant difference:
image
the only disatvantage on the time-base values is the higher noise. So what.
image

PR #320, opened 2025-05-26
Update zendurebase.py

harrymayr:
Bugfix Remaining time:
Should be None / unknown, if Time is None


FireSon:
Thanks I have updated it in my own code

Issue #272, opened 2025-05-18
More efficient hysteresis than Zendure implemented yet - specially in clusters and with high solar intake


MacdiverOne:
Is it planned (or already integrated and possibly overlooked by me) to integrate a hysteresis function into the automated power control, as the hysteresis that Zendure has integrated yet is not really effective. The existing implementation by Zendure alternates in small steps between charging and discharging specially at high solar intake and neither promotes the durability of the battery nor the efficiency of energy utilization.

fodhamburg:
Yes! Its there ✅ 

In the current pre-release version is an offset which allows to control a hysteresis on your end.
So you can a) jump on the pre-release version or wait for the next official version that will introduce it.

MacdiverOne:
I'm currently running the pre-8 which does not offer any hysteresis to be configured as far as I know.

You are not referencing the "offset to 0-grid-input-automatism" in zendurermanager.py (pull request from @harrymayr pending) we are talking about nowadays with @harrymayr and @FireSon?

harrymayr:
> Is it planned (or already integrated and possibly overlooked by me) to integrate a hysteresis function into the automated power control, as the hysteresis that Zendure has integrated yet is not really effective. The existing implementation by Zendure alternates in small steps between charging and discharging specially at high solar intake and neither promotes the durability of the battery nor the efficiency of energy utilization.

There is a delay of 10 seconds between the change from charging to discharging or vice versa. In this time the state is standby (power=0). This time is defined in const.py, so if you want to test a longer time, feel free.
It may be, that the power which come in from the PV inputs is not always taken correct into account, depending on the state of the Bypass.

Anyway, we are happy if you share some proposals, with clear definitions of input source, bypass, demand, ...

MacdiverOne:
This hysteresis request is not about short intervals. It is primarily about preventing a short collection phase (3%) and immediate switching to discharge, as Zendure currently does. Especially in situations where the solar input just about matches the load profile, this leads to a constant zig-zag between charging and discharging. If it were now possible to shift the trigger point for the next discharge, let's say to a minimum of 10-20% difference to the SOC at the start of charging (if < max SOC), this would certainly be good for the entire system. At least I think so. Possibly also relays and batteries...

Translated with DeepL.com (free version)

FireSon:
@MacdiverOne this actually one of the things were I think we can focus on now, since most of the communication en Zendure devices are now working properly, which had my main focus, in the last few weeks. 

[...]

PR #331, opened 2025-05-28
Offset

harrymayr:
This PR does the following:

- allow numbers to be persistent
- add a config entry "Experimental functions"
- add an offset-slider to the Zendure-Manager, if Experimental functions are enabled
- This offset-slider allow a threshold around 0. You can select, (offset > 0) if you want to spent more into grid and use not all of the solar power to load the battery, with the idea to get as less from the grid as possible. Or you select offset < 0 and than you be sure not to spent battery power into grid and also use all solar power to charge batteries. This is usefull in winter, where you need to get power from the grid anyway.
- This also reduces switching between charge and discharge. It create a kind of corridor where the system is in standby.
- If we now also increase TIMEIDLE (or make it also configurable) to let's say 60 or 120 sec. Then this mean that switching will reduced even more. This is not yet in this PR
- We also can consider to increase the idle-time in special circumstances, but did not think for now on it, how this can work.

This is part of the discussionin #272

### DIFF
--- a/custom_components/zendure_ha/config_flow.py
+++ b/custom_components/zendure_ha/config_flow.py
@@ -16,7 +16,7 @@ from homeassistant.helpers import selector
 
 from .api import Api
 from .const import (CONF_MQTTLOCAL, CONF_MQTTLOG, CONF_P1METER, CONF_WIFIPSW,
-                    CONF_WIFISSID, DOMAIN)
+                    CONF_WIFISSID, CONF_EXPERIMENT, DOMAIN)
 
 _LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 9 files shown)


kumy:
@harrymayr Wondering if this piece of work would allow to start charging the batteries when the solar surplus is "greater than a threshold" (ie >250W)? Goal is to feed the "low" surplus to the boiler via a PVRouter as (I read somewhere, but sorry can't find the source anymore) the efficiency is poor when charging at low power. Thanks

harrymayr:
@kumy The problem is, if you feed some power into the Boiler, there is no surplus seen on the P1 Sensor. On the other hand to feed only surplus energy into the Boiler, you may control it by using outputPackPower and packInputPower. You can start feeding in, if outputPackPower is reaching some threshold and increase the power into the Boiler to your wanted target. You need to reduce the power  into the boiler, if packInputPower start to rise.
In other words, if you build a controller outside this integration which uses the power into your boiler and the values for the power in and out of the batteries, you can fullfil your needs.

Your topic could be something for the [Discussions](https://github.com/FireSon/Zendure-HA/discussions)

BTW, here is the link to PVRouter https://openenergymonitor.org/mk2pvrouter/

FireSon:
Harry, we need to fix this in the new version, since there are some many changes, I close it for now

PR #332, opened 2025-05-29
Time update

harrymayr:
The Hyper does not provide any charging time. So calculate in any case triggered from outputPackPower


FireSon:
@harrymayr I have updated this in a slightly different way, but without the frequent updates

PR #352, opened 2025-06-05
Update zendurermanager.py

harrymayr:
Additional logging, because on the analysis of different logs, it seems, that sometimes not all originally recognized devices are in the cluster.


FireSon:
@harrymayr I have some idea's to improve the handling of 'available' devices, I prefer to add it together with that change, so you can see that x of y devices are used.

PR #364, opened 2025-06-08
Update zenduredevice.py

harrymayr:
Add batterie only, if the serial numbers are the only content of packData. This should solve @Linus86529 ghost batterie issue, where single entries with SN and data came in over MQTT.


harrymayr:
Still work in progress. I have send to Linus another Version which he test right now.
But there are 2 different commands. The 1st one with only the serial-numbers of all attached batteries and later different data with the correspondig serial.
I thought, that the info with all the serials is correct, but it isn't always the case.

FireSon:
Yes that is why I did it this way 😄

PR #377, opened 2025-06-11
Handle ble errors & mqtt reset

harrymayr:
There is no need to suppress MQTT communication in local mode if there are BLE errors and there is also no need to force a MQTT reset, if MQTT commands come in.

I have sometimes BLE timeouts, the actual implementation suppress MQTT communication and start a MQTT reset which switch off local mode. This is not necessary.

image

### DIFF
--- a/custom_components/zendure_ha/zendurermanager.py
+++ b/custom_components/zendure_ha/zendurermanager.py
@@ -332,7 +332,7 @@ class ZendureManager(DataUpdateCoordinator[int], ZendureBase):
                     if device.mqttLocal == 1:
                         device.mqttStatus()
 
-                if ZendureDevice.mqttIsLocal and (device.mqttLocal < 8 or device.mqttZenApp != datetime.min) and topics[0] == "":
+                if ZendureDevice.mqttIsLocal and ((device.mqttLocal & 15) < 8 or device.mqttZenApp != datetime.min) and topics[0] == "":
                     ZendureDevice.mqttCloud.publish(msg.topic, msg.payload)
 
             else:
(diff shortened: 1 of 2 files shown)


harrymayr:
This changes seems to provide more info for the Zendure Server (connection mode still say local). Not sure, if this is wanted and also not sure why this happen 😑.  But I find it not bad.
Screenshot_20250612_061001_Zendure

custom_components/zendure_ha/zendurermanager.py
```diff
@@ -332,7 +332,7 @@ def mqttMsgLocal(self, _client: Any, _userdata: Any, msg: Any) -> None:
                     if device.mqttLocal == 1:
                         device.mqttStatus()
 
-                if ZendureDevice.mqttIsLocal and (device.mqttLocal < 8 or device.mqttZenApp != datetime.min) and topics[0] == "":
+                if ZendureDevice.mqttIsLocal and ((device.mqttLocal & 15) < 8 or device.mqttZenApp != datetime.min) and topics[0] == "":
                     ZendureDevice.mqttCloud.publish(msg.topic, msg.payload)
```
FireSon, line 336: Nice catch 😄

FireSon:
The check_reset is not a property of the ZendureDevice, and it is used to regularly check all devices. I have added the device.mqttLocal  & 15 fix, but the rest is not possible.

harrymayr:
> The check_reset is not a property of the ZendureDevice, and it is used to regularly check all devices. I have added the device.mqttLocal & 15 fix, but the rest is not possible.

Hmm, I've added this to my code and see no error in the logs.

PR #383, opened 2025-06-12
(dis)charge only

harrymayr:
This PR add the translations for the combination with the BLE Error and add two new manager modes: Smart Charge and Discharge only. 
In fact it set the setpoint to 0 if the P1 sensor request charging if only discharging is allowed and vice versa.
I also added a change of the Manager-Mode to None, if the min/max SOC is reached in Manual or (Dis)Charge only mode.


FireSon:
I cherrypicked your PR, and implemented it slightly different.

PR #435, opened 2025-07-02
Update solarflow800Pro.py

harrymayr:
fix BatVolt
at least this device has also a timestamp which could be changed to a readable format
perhaps somethink like `self.sensor("ts","{{ value | timestamp_custom('%x %X')}}", None, "timestamp"),`? Need to add translations also.
image

### DIFF
--- a/custom_components/zendure_ha/devices/solarflow800Pro.py
+++ b/custom_components/zendure_ha/devices/solarflow800Pro.py
@@ -52,7 +52,7 @@ class SolarFlow800Pro(ZendureDevice):
         sensors = [
             self.sensor("hubState"),
             self.sensor("solarInputPower", None, "W", "power", "measurement"),
-            self.sensor("BatVolt", None, "V", "voltage", "measurement"),
+            self.sensor("BatVolt", "{{ value | int / 100 | round(1)}}", "V", "voltage", "measurement"),
             self.sensor("packInputPower", None, "W", "power", "measurement"),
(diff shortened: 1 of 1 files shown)


judie07:
Perfect, thank you very much, looks good.
Zendure-Install_7

judie07:
I have seen two other displays:
Fan speed is always at zero, and the Hyper Temp.
The connected fan works perfectly.

Zendure-Install_8

harrymayr:
> I have seen two other displays: Fan speed is always at zero, and the Hyper Temp. The connected fan works perfectly.
> 

Hyper Temp we can change, but if the Fan speed does not change, then this information is not provided over MQTT.
BTW, does Hub Status remains unknown all the time?

judie07:
@harrymayr The fan is running, but I can't say for sure whether the speed is changing. I will check with a voltmeter. Important: The SF800pro is NOT yet connected via MQTT, as there were still difficulties when integrating it. The recommendation was to go via the cloud first.

harrymayr:
> @harrymayr The fan is running, but I can't say for sure whether the speed is changing. I will check with a voltmeter. Important: The SF800pro is NOT yet connected via MQTT, as there were still difficulties when integrating it. The recommendation was to go via the cloud first.

This is also a communication with MQTT. The information is the same, regardless of cloud or local MQTT communication.

judie07:
Okay, thanks for the tip.

PR #489, opened 2025-07-27
Reflect legacy device needs

harrymayr:
Possibility to change passMode/gridReverse for legacy devices


FireSon:
Hi Harry, I have changed many things, so I skip this PR.

Issue #515, opened 2025-08-01
max inputLimit is limited to 1200W for SF2400AC with 1.1.0 pre-13


SukramJ:
The maximum inputLimit For Solarflow 2400AC can only be set to 1200 with the slider.
2400W is possible with 1.0.47 and the official APP.

harrymayr:
Does the SF 2400AC show an entitiy `chargeLimit` and is it's value 2400? If so, you can add this lines in device.py and restart HA:

                case "chargeLimit":
                    self.powerMin = -value
                    self.limitInput.update_range(0, value)

<img width="800" height="476" alt="Image" src="https://github.com/user-attachments/assets/2248b719-1b46-40d6-9992-ecceedbe8e43" />

SukramJ:
Works with `chargeMaxLimit`

```
                case "chargeMaxLimit":
                    self.powerMin = -value
                    self.limitInput.update_range(0, value)
```

Thanks a lot. Great work.

SukramJ:
Fixed by pre-14. Thanks

PR #521, opened 2025-08-02
Update device.py

harrymayr:
Update slider limitInput according to device chargeLimit (see #515)

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -129,6 +129,9 @@ class ZendureDevice(EntityDevice):
                 case "inverseMaxPower":
                     self.powerMax = value
                     self.limitOutput.update_range(0, value)
+                case "chargeLimit":
+                    self.powerMin = -value
+                    self.limitInput.update_range(0, value)
 
         return changed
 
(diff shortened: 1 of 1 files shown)


FireSon:
Well sorry, but this is not allowed. If you want to increase the output limit, they have to change it in the ZendureApp! You actually have to sign for that!! Once you modified it there, it will be possible to change the output limits. I actually added my devices to a seperate group today and did it myself 😄

harrymayr:
> Well sorry, but this is not allowed. If you want to increase the output limit, they have to change it in the ZendureApp! You actually have to sign for that!! Once you modified it there, it will be possible to change the output limits. I actually added my devices to a seperate group today and did it myself 😄

Mmh, but this is the value from the ZendureApp / Device. And it is the inputLimit what should be changed.

FireSon:
Oeps, sorry thought it ws about the output limit 😄

Johan1971:
I can change the output limiet power in the app on each 2400AC unit thats good.
But is there an way to set an input limit on a unit? in the app not.

What i now see when its in smart mode my 2400AC units goes 2400W input on solar input.
Is it possible to make this setting in the intregation so it can be set lower.

FireSon:
Cherrypicked the PR and added it. Thanks!

PR #529, opened 2025-08-02
Bypass

harrymayr:
Not all devices use gridReverse, but also passMode


FireSon:
Hi Harold, changed it in a simpeler way, can you check if it works for you as well?

harrymayr:
> Hi Harold, changed it in a simpeler way, can you check if it works for you as well?

At least for the Hyper it works, after I fixed the json error in DE.json (directly commited)

And as always, your coding capabilitys are better than mine :-D

PR #532, opened 2025-08-03
online-sensor for tracking status

harrymayr:


custom_components/zendure_ha/api.py
```diff
@@ -232,7 +232,13 @@ def mqttMsgLocal(self, client: Any, _userdata: Any, msg: Any) -> None:
             deviceId = topics[2]
 
             if (device := self.devices.get(deviceId, None)) is not None:
-                payload = json.loads(msg.payload.decode())
+                try:
```
FireSon, line 235: Why do we need this? Never seen this "" problem? Did you?

custom_components/zendure_ha/entity.py
```diff
@@ -186,11 +187,21 @@ def entityUpdate(self, key: Any, value: Any) -> bool:
                 entity = ZendureSensor(self, key)
 
             if entity is not None and entity.platform is not None:
-                entity.update_value(value)
+                # fix typecast error in AB3000X
```
FireSon, line 190: Same here, does this happen often? Never seen it myself

custom_components/zendure_ha/manager.py
```diff
@@ -298,6 +298,10 @@ def update_operation(self, entity: ZendureSelect, _operation: Any) -> None:
         if self.operation != SmartMode.MATCHING and len(self.devices) > 0:
             for d in self.devices:
                 d.power_set(ManagerState.IDLE, 0)
+        if self.operation != SmartMode.NONE and len(self.devices) > 0:
```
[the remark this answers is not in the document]
harrymayr: It's not when using OFF, but using the Smart Modes. If the limits are not 0, I sometimes see a wird behavior (stopping charging at a Rondo limit, ...)

custom_components/zendure_ha/entity.py
FireSon: Nice, I could see it as well on my SF2400 😄 will fix that!

FireSon:
Hallo Harry, the last couple of days have been quite busy with finishing up version 1.1.0! As far as I can see, all the issues from your PR have been addressed, except for the setting of the output limit, because it was not 100% clear to me. 
Would you be OK with testing the new version, and create a new PR if things are not working properly? Would be, as allways, much appriceated!!
Will close this PR for now...

harrymayr:
> Hallo Harry, the last couple of days have been quite busy with finishing up version 1.1.0! As far as I can see, all the issues from your PR have been addressed, except for the setting of the output limit, because it was not 100% clear to me. Would you be OK with testing the new version, and create a new PR if things are not working properly? Would be, as allways, much appriceated!! Will close this PR for now...

That's OK. I'm testing out the moment with random output- and inputlimits using smart matching and had no problems so far.

PR #594, opened 2025-08-08
Missing entities added.  German translation updated.

z-master42:
Missing entities added.

German translation updated. Missing translation added, existing translation updated.

### DIFF
--- a/custom_components/zendure_ha/entity.py
+++ b/custom_components/zendure_ha/entity.py
@@ -57,6 +57,7 @@ class EntityDevice:
         "outputHomePower": ("W", "power"),
         "gridInputPower": ("W", "power"),
         "gridOffPower": ("W", "power"),
+        "gridPower": ("W", "power"),
         "solarInputPower": ("W", "power"),
         "solarPower1": ("W", "power"),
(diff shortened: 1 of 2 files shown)


custom_components/zendure_ha/entity.py
```diff
@@ -100,12 +101,14 @@ class EntityDevice:
         "ambientLightColor": ("none"),
         "ambientLightMode": ("none"),
         "ambientSwitch": ("none"),
-        "PowerCycle": ("none"),
-        "packInputPowerCycle": ("none"),
-        "outputPackPowerCycle": ("none"),
-        "outputHomePowerCycle": ("none"),
+        "PowerCycle": ("W", "power"),
```
FireSon, line 104: The entities I have marked with "none", are not added to HA because they are for internal use by Zendure. Is there a reason you want to change this?

z-master42: No does not have to. I also don't know how long the cycle is. But they also appear with “none” in the entity list and are therefore no longer really internal. So if they appear, you can also give them the right entity ;-)

PR #689, opened 2025-08-21
Fix powerAct calculation: use outputHomePower

Zanooon97:
use outputHomePower to add a "backup" value for the Actualpower for eatch device

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -380,9 +380,13 @@ class ZendureDevice(EntityDevice):
 
     async def power_get(self) -> int:
         """Get the current power."""
-        if not self.online or self.outputHomePower.state is None or self.gridInputPower.state is None:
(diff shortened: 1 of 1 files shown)


FireSon:
@Zanooon97 Sorry I did not have time to look into this change, if I understand correctly you wanted to use outputHomePower? Is that resolved now? 😄

Zanooon97:
@FireSon Not resolved. I made it for me like my pull request, i want only show you that it works. Make it in your way. 
The actual power stucks at 0watts when battery is inactive (maxsoc) so check the power_get function

Zanooon97:
bye

Zanooon97:
without adding output home power its not working see screenshot
with adding output home power is working see screenshot
you also need to see that the solarinput is higher than the real output when its going in bypass, so solarinput cant be used!

immagine
immagine

PR #718, opened 2025-08-27
Update device.py

harrymayr:
manual change of minSoc will keep changed == false and electricLevel could be longer time the same


FireSon:
Hi Harry, fixed it a little bit different but thanks 😄

PR #727, opened 2025-08-30
Corrected fr.json

Youpimatin:


Youpimatin:
maybe a stupid question ahead with an easy reply :
I would like to add traductions for some items that are not in any .json
for example "charge_max_limit", "grid_off_power" and others.
Where can I find entities names to use ? 

Even more :
Where can I find more information about these same entities ?
For example, is "charge_max_limit" the maximum charge input into the AB3000X, or the total maximum AC input for battery charging AND off-grid power ? something else ? Don't know if I'm clear, but to have the right notions and the right words in the traductions.

PR #740, opened 2025-09-05
Update manager.py

harrymayr:
Don't use more power on Startup as requested


FireSon:
Hi Harry, with all the recent changes can you check if all is covered?

PR #771, opened 2025-09-12
Update Onlysolar, add min charging window for battery

Zanooon97:


FireSon:
Why would you change these numbers? They are not related to socLimit values...

Zanooon97:
I lost the other files — sorry.
Also, my manager stopped working since you switched to using packinput for the calculation.
And I still haven’t understood what you want to do with the d.usesolar attribute.

New: DeviceState.MIN_SOC_CHARGE_WINDOW:
This window overrides the solar logic.
Devices in MIN_SOC_CHARGE_WINDOW are sorted to the end of the queue.
If surplus power is available, it’s used; in this phase the battery is never discharged—only charged. The household load has priority.

Configuration:
In const.py, SmartMode.MIN_SOC_CHARGE_WINDOW is set to 10% of the window.

“Solar-only” mode:
I changed how devices are sorted and balanced. Sorting purely by solar input caused constant switching because power dips briefly when outputs turn on/off.
It only became stable when both devices were used in parallel. An active device can switch fast enough that it goes unnoticed.
Weighting is based on the remaining energy to charge (kWh) so all devices—despite different solar inputs—converge toward 100%.

Idea per-device slider entity to set MIN_SOC_CHARGE_WINDOW would likely be useful.

When there’s enough solar energy to charge the battery and supply the house,
power = powerDischarge - powerCharge + p1 effectively collapses to just p1.
As a result, powerUpdate only sees p1, which can be 0. When it’s 0, all devices get disabled.

If gridReverse is off, the solar input is simply missing from all calculations — even when bypass is enabled.
What do you want to accomplish with useSolar and bypass?
Best regeards

FireSon:
@Zanooon97 It has been a bit hectic, with a lot of changes in the last days. I am sorry that your PR was crushed between all of this.
The SolarOnly option is now basically implemented in the current version. 
The idea of the MIN_SOC_CHARGE_WINDOW  I did not really understand, so maybe you can explain it a bit more.
Are there any other changes in your PR which I missed?

Zanooon97:
@FireSon Ja ich habe das alles Fetig gemacht und Hier Erklärt:
https://github.com/Zendure/Zendure-HA/pull/782

PR #782, opened 2025-09-15
Personalisierte Batterie Ladefenster

Zanooon97:
Personalisierte Ladefenster (relativ zum Min-SoC)
Zwei Fenster steuern strikt das Verhalten zwischen Tiefentladung und Schonbetrieb.

Fenster 1 – Priorisiertes Schnellladen
Aktivierung: sofort bei Erreichen von Min-SoC.
Bereich: Min-SoC → Min-SoC + Fenster 1.
Verhalten: Batterie gilt als praktisch leer und wird mit maximal möglicher Leistung geladen.
Zweck: Schneller Ausstieg aus Tiefentladung.

Fenster 2 – Haus priorisieren (Batterieschonung)
Umschaltung: ab Min-SoC + Fenster 1 (nur wenn Fenster 2 > 0).
Bereich: (Min-SoC + Fenster 1) → (Min-SoC + Fenster 1 + Fenster 2).
Verhalten: Keine Entladung der Batterie; Laden nur mit Überschussenergie.
Zweck: Zyklen vermeiden, besonders bei schlechtem Wetter.

Steuerung
Zwei Slider pro Gerät.
Slider = 0 → Fenster aus.
Beide = 0 → Standardverhalten wie bisher.
Binary-Sensoren zeigen aktives Fenster.

Formel
Obergrenze Fenster 1 = Min-SoC + Fenster 1
Obergrenze Fenster 2 = Min-SoC + Fenster 1 + Fenster 2

Beispiel
Min-SoC: 5 %
Fenster 1: 10 % → aktiv 5 %–15 %
Fenster 2: 10 % → aktiv 15 %–25 %

Merke: Fenster-Prozente sind Aufschläge über Min-SoC → Min-SoC + Fenster 1 + Fenster 2.

Zendure-Manager Ergänzungen
Power-Entitäten: Standardabweichung, P1-Durchschnitt, Power-Durchschnitt (für HA-Diagramme).

Batterie-Icons an Min/Max-SoC-Steuerelementen.

Screenshot_2025-09-15-20-53-00-030_io homeassistant companion android

### DIFF
--- a/custom_components/zendure_ha/const.py
+++ b/custom_components/zendure_ha/const.py
@@ -32,6 +32,7 @@ class DeviceState(Enum):
     INACTIVE = 3
     STARTING = 4
     ACTIVE = 5
+    MIN_SOC_CHARGE_WINDOW = 6
 
 
 class ManagerState(Enum):
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -21,7 +21,7 @@ from .binary_sensor import ZendureBinarySensor
 from .button import ZendureButton
 from .const import DeviceState, SmartMode
 from .entity import EntityDevice, EntityZendure
-from .number import ZendureNumber
+from .number import ZendureRestoreNumber, ZendureNumber
 from .select import ZendureRestoreSelect, ZendureSelect
 from .sensor import ZendureRestoreSensor, ZendureSensor
 
@@ -108,6 +108,7 @@ class ZendureDevice(EntityDevice):
         self.actualKwh: float = 0.0
         self.activeKwh: float = 0.0
         self.state: DeviceState = DeviceState.OFFLINE
+        self.min_soc_charge_window_ready = False
 
         self.create_entities()
 
@@ -115,12 +116,14 @@ class ZendureDevice(EntityDevice):
         """Create the device entities."""
         self.limitOutput = ZendureNumber(self, "outputLimit", self.entityWrite, None, "W", "power", 800, 0, NumberMode.SLIDER)
         self.limitInput = ZendureNumber(self, "inputLimit", self.entityWrite, None, "W", "power", 1200, 0, NumberMode.SLIDER)
-        self.minSoc = ZendureNumber(self, "minSoc", self.entityWrite, None, "%", "soc", 100, 0, NumberMode.SLIDER, 10)
-        self.socSet = ZendureNumber(self, "socSet", self.entityWrite, None, "%", "soc", 100, 0, NumberMode.SLIDER, 10)
+        self.minSoc = ZendureNumber(self, "minSoc", self.entityWrite, None, "%", "soc", 100, 0, NumberMode.SLIDER, 10, icon="mdi:battery-10")
+        self.socSet = ZendureNumber(self, "socSet", self.entityWrite, None, "%", "soc", 100, 0, NumberMode.SLIDER, 10, icon="mdi:battery")
         self.socStatus = ZendureSensor(self, "socStatus", state=0)
         self.socLimit = ZendureSensor(self, "socLimit", state=0)
         self.byPass = ZendureBinarySensor(self, "pass")
         self.gridReverse = ZendureSelect(self, "gridReverse", {0: "disabled", 1: "allow", 2: "forbidden"}, self.entityWrite, 0)
+        self.MSCW1 = ZendureRestoreNumber(self, "MSCW1", self.entityWrite, None, "%", "soc", 20, 0, NumberMode.SLIDER, 10, icon="mdi:battery-10")
+        self.MSCW2 = ZendureRestoreNumber(self, "MSCW2", self.entityWrite, None, "%", "soc", 20, 0, NumberMode.SLIDER, 10, icon="mdi:battery-10")
 
         fuseGroups = {0: "unused", 1: "owncircuit", 2: "group800", 3: "group1200", 4: "group2000", 5: "group2400", 6: "group3600"}
         self.fuseGroup = ZendureRestoreSelect(self, "fuseGroup", fuseGroups, None)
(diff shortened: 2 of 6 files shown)


FireSon:
Hi @Zanooon97 personally I am a fan of keeping things as simple as possible especially since we have more than 1000 users at this moment.

To be honest, I do not really understand why these windows are necessary? What is the gain? And why do you need to set these limits individually? We have minSoc and socSet to prevent the battery from unloading too much. 
Would using the smartmode ChargeOnly not do the same thing? Also for people who have an existing PV system?

The other problem is that we are in the process of releasing version 1.1.4, and I do not want to have conflicting changes to make testing more difficult. And I see a lot of changes in things I have introduced in my latest changes.

Zanooon97:
I get your point about keeping things simple. The main reason I added the windows is to cover some specific scenarios that the existing minSoc/socSet or ChargeOnly can’t handle.

For example, with the windows I can prioritize the house first – e.g. in window 2 the logic ensures that PV is used for the house load before charging the battery. This gives users finer control over how surplus energy is distributed, preventing the battery to charge 200watts and few houer later discharge with 800watts 2% to minsoc.

Also, a couple of users have already asked for this kind of flexibility, so I believe there is real demand for it.

Idk for me the hous priority at low stage is working better and has less on off

cybe:
While I appreciate the flexibility that comes with this proposal, I tend to agree with @FireSon that this might be too complicated for general usage:

- Do I need to specify these values? What are sane values?
- Are those window minimums percentage points or percentages? (It's the former)
- Can it be, that those window minimums mark a range on the SoC scale of 0-100%? (Nope, but at least I myself thought so at first)

As I understand, the proposed change corresponds to `MSCW1=0%` and `MSCW2=2%`  in `1.1.4-pre6`, which is hardwired. I'm generally happy with this configuration personally, but understand other needs as well.

I too think it's viable to approach such needs with different manager operation modes in combination with home assistant automation. Like setting it to ChargeOnly for desired SoCs. Given that those operation modes do work.

Zanooon97:
@cybe 

If I have two devices with PV modules oriented east and west, it becomes much more complicated – almost impossible – to find the right configuration using only the different modes.

These windows should be understood as percentage points added to minSoc. 
For example, if your minSoc is set to 10%, then your device enters SocEmpty. 
If window 1 is set to 10%, the device remains marked as SocEmpty and will continue to charge with maximum available power until 20%. 
If window 2 is set to 10%, then the device is marked as being in bypass mode until 30% (the description could be improved). In this mode, solar power is directed to the house, and if solar is not sufficient, the rest is taken from the grid to prevent the battery from discharging. Only the surplus goes into the battery.

The advantage is that the device won’t constantly charge and then discharge again after only 2% of charging. Once the SoC goes above 30%, the device leaves these stages and becomes fully available again. 

I added window 1 because there might be users who discharge deeper than 10%. In that case, the battery should be brought out of this state as quickly as possible.

The windows are admittedly not well explained and not obvious at first glance – but they are meaningful. 

Personally, I have the first window disabled (set to 0) and the second set to 10%. This way, in bad weather, I avoid the constant cycle of charging and then discharging.

cybe:
@Zanooon97 thank you for illustration the purpose further. 

I'm in the same boat as you, as I would like to avoid constant cycling also. And I acknowledge that there probably isn't one hardwired number that would fit every setup there is, so making window2 configurable seems reasonable to me. Maybe we can find a better name for it, but this feature is a 👍 from me.

As for window1: I personally wouldn't use this feature. I just set the `min_soc` to my preferred value and I'm fine with the current soc just sitting there, even if there is solar power available. But as some people like to use this feature I won't vote on this.

MacdiverOne:
Just a feet in the open door to active steering as the shadows get longer and the snow comes into sight.

For me it seems, that there are ways to use w1 and w2 (there must be better names out there for the functions and the UI) for calculated reactions (and preparations) to actual and proposed weather conditions (sun/snow/growing+shrinking shadow areas as the sun moves into winter and summer elevation/clouds/whatsoever)

To make it "consumer-ready": the AI that could be predictive and clever "HACSZIAIAIAI" ;-)

Maybe the first which get's it done to manage the gap between the price of stored kWh and the time to release the stored kWh if the interrest is appropriate (and not, if the actual price get 1ct higher than the low-price-limit) and more: with an eye on the respective grid meter...

Just dreaming...

cybe:
Yeah, AI could be the way to go moving forward in the long run ;)

As for the time being, I'm more and more on board with having a more fine grained control as proposed. Maybe this could be declared in the UI as some kind of advanced feature? In addition, maybe we can improve the wording for the concept or even the way to configure it.

But I find it very difficult to describe or even implement the concept more user friendly without clashing with `min_soc` . Both windows seem similar, but are very different conceptually: The first one feeds all solar power to the battery up until a certain soc. The second one feeds solar power to the home and excess power to the battery, but won't use the battery to feed the home up until a certain soc.

Isn't the goal an automatic switch of the manager operation based on SoC?
* When SoC is below `min_soc + window1`: `off`
* When SoC is above `min_soc + window1` and below `min_soc + window1 + window2`: `smart_charging`
* When SoC is above `min_soc + window1 + window2`: `smart`

FireSon:
@Zanooon97 Thanks for the extra explanation, I think I understand it better what happens in your setup. Since I have a seperate EV system, all my batteries are charged and discharged equally. But since your solar panels are connected directly to a Zendure device, some of them are charged quicker than the others, due to the orientation of the solar panels right?

In my latest commit, which is a work in progress 😄, I have added an extra "strategySolar" which should handle the situation where the solar power produced by 'Zendure' is bigger than the house demand. I think this is also a place where we could solve your specific situation. For the 'strategyCharge', which is used when the house has a surplus.
As an example, I added the option to give priority to devices for instance if they have a electricLevel < SmartMode.SOCMIN_OPTIMAL (=22). In the strategyCharge we could add the possibility to set one device to charging and another to discharging if the latter has a solar surplus.  In the end this could lead to an equal charging of all batteries.

Please let know what you think!

Zanooon97:
immagine
immagine
immagine
Hi @FireSon 
With the current version, the charge/discharge cycle works like this: once soclimit 2 is reached, the battery starts charging. But when soclimit 0 is reached, the battery discharges at full force — which happened several times today while I was cooking.

I modified the function a bit and also renamed it to make it more self-explanatory!
Without such a change, I would otherwise have to pull a device out of the fuse group and rely on HA automations, which quickly becomes overly complex.

I’m only using Zendure devices with PV modules and no external PV system, so this should not cause any issues or conflicts for purely AC-connected devices either. We could even make it appear automatically only when PV panels are actually connected.

I’ll never use the strategyCharge if I don’t have an external PV system.
I get what you had in mind though, hehe.

PR #805, opened 2025-09-21
Update manager.py

harrymayr:
wrt to my simulations, this solution does not use batterie-power in case there is enough solar power. Also switching between devices is (mostly) reduced.


cybe:
I second this. Also using solar power directly and not charging the battery. We already had this behavior in 1.1.4-pre7. But not anymore since [1e3bbbc](https://github.com/Zendure/Zendure-HA/commit/1e3bbbcd0fd3d3e2f7082a759908784351e90f92):

image

SoC min is set to 20%. The low solar power is used to charge the battery. But it should be fed to the house while enabling the bypass.

Zanooon97:
@cybe this is what i am doing the hole time and nobody is understanding hahaha
https://github.com/Zendure/Zendure-HA/pull/807

cybe:
> @cybe this is what i am doing the hole time and nobody is understanding hahaha #807

Yes, I think this should be the default behavior. But it seems to be a bit niche im wondering? I'm not sure why it was implemented as part of 1.1.4-pre5 but lost since:

> * Solar matching, when the house requires power, first the available solar power is used

PR #807, opened 2025-09-22
Extend Charge/Discharge Cycle

Zanooon97:
Her you see how it is now. 
And also solved a little bug in solarstrategy when delta p1_set / solar is to low then it is using battery energy instead of using the next device solarpower.

immagine
immagine
immagine
immagine
immagine


cybe:
I like your approach in having a priority setting 👍

Zanooon97:
Minimum Charge / Discharge Cycle (%) 
Minimum Charge / Discharge Cycle (Priority) 
is a better name

FireSon:
Hi @Zanooon97 thanks for the work. Unfortunately I cannot complete this PR, as there are too many changes in one PR. It also contains a lot of thing I do not want to add, like a lot of new entities, which you probably use for testing. Another thing is that you also change things I am working on myself, which does not help. And adding new functionality like home/battery priorities, would be better to discuss first.  
I really like your input, and I think you have a point with (most) of the proposed changes! But I want to keep the direction in which we develop the Integration under control, and as 'simple' as possible (as I have said before). So the first step is to fix the power distribution and once that is stable (version 1.1.4) I am open to new suggestions/changes...
Hope you understand the situation!

auzbb:
Hi @Zanooon97 I installed your version and currently have 12% SoC and 30W power (SF800Pro). I would expect the bypass to be activated and the power to go into the house (I have set priority to the house). However, it is only charged into the battery, regardless of whether I have the charge/discharge cycle extension (%) set to 1% or 10%. Am I missing something? I want the house to be supplied first when there is little PV power, and only when the battery falls below minSoC or there is more PV power available than is needed in the house, I want the battery to be charged. And the battery should only start drawing power from the battery when the charge/discharge cycle extension (%) is set to 10% at, for example, (if minsoc is set to 10) at 20%. That's the goal of your adjustment, right?

cybe:
> ... have 12% SoC and 30W power (SF800Pro). I would expect the bypass to be activated and the power to go into the house ...

Most likely because of https://github.com/Zendure/Zendure-HA/blob/46555bf661c9cd86935702cafcd9720118ef66ba/custom_components/zendure_ha/const.py#L52

Bypass is only activated for solar power > 50 Watts, because of losses of the inverter itself.

Also see https://github.com/Zendure/Zendure-HA/discussions/792#discussioncomment-14424618

Zanooon97:
the bypass is activated at 50watt and turned off at 30

> Hi @Zanooon97 I installed your version and currently have 12% SoC and 30W power (SF800Pro). I would expect the bypass to be activated and the power to go into the house (I have set priority to the house). However, it is only charged into the battery, regardless of whether I have the charge/discharge cycle extension (%) set to 1% or 10%. Am I missing something? I want the house to be supplied first when there is little PV power, and only when the battery falls below minSoC or there is more PV power available than is needed in the house, I want the battery to be charged. And the battery should only start drawing power from the battery when the charge/discharge cycle extension (%) is set to 10% at, for example, (if minsoc is set to 10) at 20%. That's the goal of your adjustment, right?

you right, you need to have over 50 w to activate the bypass

auzbb:
Bildschirmfoto 2025-09-24 um 17 21 00

Well I understand that explanation - but then why the bypass is switching so often (multiple times per minute)?

PR #823, opened 2025-09-28
Update manager.py

harrymayr:
Fast start of charging, if all batteries are empty


FireSon:
Hallo Harry, I have added a peak detection for the average value, can you have a look at that? should fix your problem...

PR #855, opened 2025-10-21
Fix and add aggr total sensors

tmhrzgr:
Update for multiple files:

- Fix for existing offGrid and aggroffGrid entities: corrected case sensitivity and bound entities to the value “gridOffPower” provided by the Zendure device → offGrid represents the output power of the Offgrid socket, aggrOffGrid represents the total kWh counter.
- Added aggrHomeInput entity → combines HomeInput for input power from AC and input power from the Offgrid socket, and aggrHomeInput for the total kWh counter.
- Added friendly names in English and German.

Result should solve my own feature-Request 🥳 #847 

Note: successfully tested with the SolarFlow 800 Pro. I wasn’t able to test Offgrid input from another inverter due to the lack of a second system. My expectation in that scenario is that the homeInput value will be positive when power is supplied from Offgrid. If anyone can test this, that would be fantastic :)

### DIFF
--- a/custom_components/zendure_ha/translations/en.json
+++ b/custom_components/zendure_ha/translations/en.json
@@ -114,6 +114,18 @@
       "aggr_solar_total": {
         "name": "Total Solar"
       },
+      "aggr_grid_input_power_total": {
+        "name": "Total from Grid/Offgrid"
+      },
+      "aggr_grid_off_power_total": {
+        "name": "Total used by offgrid"
+      },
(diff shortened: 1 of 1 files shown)


tmhrzgr:
device.py-update-description is wrong, didn't needed to fix offGrid here - just added the aggr entities :). 

Maybe for a future update: the complete offGrid entities should only be provided in the individual device files as not every device can use AC charging and have a Offgrid socket.

Issue #801, opened 2025-09-20
AB2000X


Spottttt:
I have added AB2000X to my two hypers and battery type is not good :       

<img width="317" height="176" alt="Image" src="https://github.com/user-attachments/assets/d77e4f91-3550-464c-8fa2-df9c574f529a" />

So , i change  device.py :   

            case "C":
                model = "AB2000" + ("S" if sn[3] == "F" else "X" if sn[3] == "E" else "")

And now it is good:   

<img width="317" height="178" alt="Image" src="https://github.com/user-attachments/assets/88d86571-095c-4638-8b76-573b1b415691" />

But micrologiciel become bad.

So is it possible to add AB2000X ?

Zanooon97:
case "C":
            model = "AB2000" + ("S" if sn[3] == "F" else "X")

?

tmhrzgr:
Hi @Spottttt 

my SolarFlow 800 Pro have a integrated AB2000X and the fourth value of my serial number is a "E" as well - so I think this identifier for the device should be correct.

In my instance the AB2000X firmware version isn't shown in integration version 1.1.3 at all. With 1.1.4-pre7 it is shown correct. I've changed the code for the device identification in device.py in my instance and it works as it should with the correct device name.

I've created the PR #857 to identify the device correctly. 😊

PR #857, opened 2025-10-22
Added AB2000X model identifier

tmhrzgr:
Added identification for AB2000X. I can confirm the "E" as the fourth position of the serial number with the SolarFlow 800 Pro integrated AB2000X and in issue #801 the E is confirmed as well.


Spottttt:
@tmhrzgr   
Hello
This is exactly what I wanted to do...
Thank you.

PR #871, opened 2025-11-02
Fix/p1 meter unit conversion

swissky:
Fix: Add automatic P1 meter unit conversion (W/kW)
Adds support for P1 meters that report in kilowatts instead of watts.

Problem:
- Many P1 meters report power in kW (e.g. 2.5 kW)
- Code assumed all values were in Watts
- This caused smart mode to not work (values 1000x too small)
- IGNORE_DELTA and thresholds were ineffective

Solution:
- Automatically detect unit from sensor attributes
- Convert kW to W when necessary (multiply by 1000)
- Log detected unit on first update for debugging
- Warn if unit is unknown
- Support multiple unit variations (kW/kilowatt/kilowatts, W/watt/watts)

Examples:
- P1 sensor shows '2.5 kW' → Converts to 2500 W ✓
- P1 sensor shows '500 W' → Uses as-is (500 W) ✓
- P1 sensor shows '-1.2 kW' → Converts to -1200 W ✓

Benefits:
✅ Works with both W and kW P1 meters
✅ No manual configuration needed
✅ Clear logging of detected unit
✅ Prevents smart mode from failing silently

Testing:
- Tested with kW sensor: Conversion works correctly
- Tested with W sensor: No conversion, works as before
- Log shows: 'P1 meter unit detected: kW (sensor: sensor.power_actual)'

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -66,6 +66,7 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         self.last_mode: str = "idle"  # "idle", "charging", "discharging"
         self.last_mode_change: datetime = datetime.min
         self.p1meterEvent: Callable[[], None] | None = None
+        self.p1_unit_logged: bool = False  # Track if we've logged the P1 meter unit
         self.update_count = 0
 
     async def loadDevices(self) -> None:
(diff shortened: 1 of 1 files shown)


FireSon:
@swissky Thanks for creating this PR, much appriciated. I did some cherrypicking from it and have pushed these changes into the main branch. I modified your KW solution in a slightly different way, by checking once what the UOM is and adding a p1_factor which is used each time the p1 value changes. Can you check if this works?
Some things I did not include onpurpose, since we are working on that. Especially the switching between the several modes. Once the next release is out, it might be good to revisit the ideas again, and if necessary, include them as well.

I also noticed that you modified all the _LOGGER calls, and changed them int %s  f.e.
` _LOGGER.error("Failed to decode payload encoding from device %s: %s", deviceId, err)`
Is there any special reason for that? I know that some loggers can do better queries if they store the parameters sperately, but I am not sure if this logger has that functionality...

I will close this PR for now, but most things should be added already 😄

Issue #873, opened 2025-11-04
[Bug] Battery at 100% breaks smart mode


zaubara:
### Home Assistant Version

2025.10.4

### Zendure Integration Version

1.1.4-pre7

### Describe the bug

Whenever the battery reaches 100%, the Hyper 2000 reduces its output to ~80W, although (much) more solar would be available. When in smart mode, I'd expect it that it would try to keep the house at 0W, maybe even discharge the battery - but neither works.

What's weird is that the hyper sensor seems to "see" the usage, but isn't trying to match it.

When setting the manager to off, which seems to be the only way to get more than 80W out of the Hyper, the sensors lowers its apparent "house usage" (sensor.zendure_manager_power) - although we're manually injecting 800W into the house.
The sensor seems to have problems with off-mode: the same happens when charging the battery via AC (manager off, but ac direction set to input): the battery charge isn't added to the "house usage" sensor .

<img width="482" height="374" alt="Image" src="https://github.com/user-attachments/assets/c05d7b01-4039-4d10-a7d3-8f5b946cd006" />

<img width="518" height="444" alt="Image" src="https://github.com/user-attachments/assets/c27e5ee8-e4d5-4323-8bd8-f61149925c1a" />

<img width="351" height="550" alt="Image" src="https://github.com/user-attachments/assets/d68e5c2e-aaf6-4881-9049-56410607190b" />

<img width="332" height="543" alt="Image" src="https://github.com/user-attachments/assets/652b0fd4-0f59-4f12-8e72-5936638a75ff" />

### To Reproduce

Use smart mode
Have a sunny day
Let it charge to 100%
See the problem

### Expected Behaviour

Try to match the house usage, even when the battery is full.
I'd expect it to enable the bypass mode, but keep it within the house limit: bypass-mode is set to "forbidden", which is a misnomer; bypass mode is expected to be set, but it should not inject anything back into the grid (keeping the house usage at 0). The sensor is correctly named select.hyper_2000_grid_reverse.

### What device are you using?

Hyper2000

### Log Extracts

```shell

```

PR #874, opened 2025-11-04
allow solar passthrough, even if battery is full

zaubara:
This change tries to balance the power consumption, even if the battery is full. Tested with only one device and only for a few minutes - this might create other issues down the line, and does not include any smoothing - there is probably a better way to do this.

### DIFF
--- a/custom_components/zendure_ha/fusegroup.py
+++ b/custom_components/zendure_ha/fusegroup.py
@@ -49,7 +49,7 @@ class FuseGroup:
         if solarOnly and -d.pwr_produced < d.dischargeStart:
             return 0
 
-        d.pwr = d.dischargeStart if d.state != DeviceState.SOCFULL else max(d.dischargeStart, -d.pwr_produced)
+        d.pwr = d.dischargeStart
         if len(self.devices) == 1:
             d.pwr = min(d.pwr, self.maxpower, d.dischargeLimit)
         else:
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -382,19 +382,18 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         self.power.update_value(pwr_home + pwr_battery)
         self.availableKwh.update_value(availEnergy)
 
-        # Get the setpoint adjust for possible bypass situation
+        # Get the setpoint and do peak detection
         pwr_setpoint = pwr_home + p1
-        if pwr_bypass != 0 and pwr_setpoint < pwr_bypass:
-            pwr_setpoint -= pwr_bypass
-
-        # Do peak detection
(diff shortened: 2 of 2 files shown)


zaubara:
With this change alone it's possible for the solar input to overshoot the power target to reach 0.

tmhrzgr:
Isn't setting the bypass-mode to allow/forbidden controlling exactly this scenario? allow = put everything into the grid what solar provides / forbidden: control the output of the inverter to the p1 meter?

zaubara:
In short, yes, exactly - and currently, this doesn't work when the battery is full (#873). When the battery is full, depending on the exact smart setting chosen, either none - 0 - or "some" (maybe 80W) of the solar power is inserted into the local house grid, even though the P1 would request some power and the panels would be able to deliver.

In the (even) longer form: Yes. Well, kind of - unlucky choice of words in my commit message. The Hyper doesn't seem to have a controllable bypass per se, the only thing close to it is the grid reverse setting.
But you're exactly right: enabling the grid reverse allows the Hyper to put everything in the grid that comes from the solar panels; it's irrelevant if the house uses some or all of it.
Disabling grid reverse (forbidden) should make it listen to the p1 meter, but it doesn't. power_discharge(0) disables any of the available power reaching the house, even though the house draws, say, 300W from the grid. You can see the power available plummet as soon as the battery is fully charged.

This simple change seems to have fixed the usage with the "smart_charge" setting for me. With "smart", there is too much fed into the local grid, and therefore exporting some of the energy, as its not properly capped at the requested usage. I hope I can check tomorrow if a secondary change fixes that as well.

harrymayr:
> Disabling grid reverse (forbidden)

In my opinion this means, that the inverter itself drive down it's solar output to 0, if the battery is full. This would be out of our control. Perhaps we could take out then power from the battery, but may be difficult for regulation, if solar-power switches on and off.

And your proposal will not have a different effect. Discharge(0) will not suppress the output to the house, if bypass is on. In this case it's the same as Discharge(pwr_produced)

zaubara:
> > Disabling grid reverse (forbidden)
> 
> In my opinion this means, that the inverter itself drive down it's solar output to 0, if the battery is full. This would be out of our control. Perhaps we could take out then power from the battery, but may be difficult for regulation, if solar-power switches on and off.

I disagree.
If there's demand from the P1 meter and there's enough solar power available, why would anyone want the supplied power drop to 0?
We want the power demand to be met  - in "smart" by any means (ie. the battery), in "smart_charging" by solar only - whatever the state of the battery.

> And your proposal will not have a different effect. Discharge(0) will not suppress the output to the house, if bypass is on. In this case it's the same as Discharge(pwr_produced)

Yes, but only IF grid reverse is enabled, and I'm specifically asking for disabled grid reverse.

PR #894, opened 2025-11-13
Update power calculations for home and produced power

harrymayr:
Adjust power calculations to account for off-grid conditions.

Any offGridPower taken from the device will be taken from grid (P1 increases), **if SoC is on minSoc** --> pwr_home will be negative (.. -self.homeInput.asInt) and reduce P1. This will not add the additional need on the setpoint for other devices.
So we need to substract offGridPower that pwr_home = 0 for this device, so that the additional need can be handled from other devices.

If there is an inverter on the offGrid port, the pwr_produced needs to be changed.

### DIFF
--- a/custom_components/zendure_ha/fusegroup.py
+++ b/custom_components/zendure_ha/fusegroup.py
@@ -27,7 +27,7 @@ class FuseGroup:
         if self.initPower:
             self.initPower = False
             if len(self.devices) == 1:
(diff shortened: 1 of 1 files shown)


harrymayr:
Discovered that there is too much charge, if the device with the off grid load do not charge. Will try to figure out a solution the next days.

PR #899, opened 2025-11-14
Fuse groups

harrymayr:
See comments for the single commits.

It fixes a too small output of a fuseGroup and balance the powerdistribution inside a fuseGroup on higher setpoints (larger than the dischargeLimit of a fuseGroup)


FireSon:
Fixed the poweroff when the fusegroup is unknown in the latest commit

PR #903, opened 2025-11-16
Valid HEMS status for all devices and HEMS switch for Hyper 2000

harrymayr:
The entity hems_state is not a valid indicator, if HEMS is enabled or not. But if HEMS is enabled, there is a topic "properties/energy" in the MQTT messages, which could be detected.

Maybe this works only for SF 800pro and SN 2400AC if mqtt control is setup in the App


harrymayr:
Created a new PR

PR #923, opened 2025-11-29
Take offgrid Power into account

harrymayr:


FireSon:
Added this already, without the:
 await d.power_charge(-SmartMode.POWER_START-d.pwr_offgrid)

PR #939, opened 2025-12-07
P1 sensor supports smart meter using HTTP GET requests

harrymayr:
This PR allow to use an IP address in the configuration of the P1 sensor.
This sensor can be a HA sensor or the IP address of a Zendure 3CT, Shelly 3EM, Shelly 3EM Pro or Shelly 3EM-63.
The communication with these smart meters are direct HTTP GET commands and will be polled with a refresh rate of 1 second.
The P1 sensor is shown in the Zendure Manager

### DIFF
--- a/custom_components/zendure_ha/__init__.py
+++ b/custom_components/zendure_ha/__init__.py
@@ -30,7 +30,7 @@ async def update_listener(_hass: HomeAssistant, entry: ZendureConfigEntry) -> No
     _LOGGER.debug("Updating Zendure config entry: %s", entry.entry_id)
     Api.mqttLogging = entry.data.get(CONF_MQTTLOG, False)
     ZendureManager.simulation = entry.data.get(CONF_SIM, False)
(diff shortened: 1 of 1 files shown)


CWILDMOUNTAIN:
amazing feature! works perfectly with the shelly pro 3EM

vouagnet:
Hello All,

Could you confirm if the PR #939 is inclued in the last official release (1.2.5) ?
If not, it is planned for the next one ?
I think it will be really appreciate, to win in reactivity with the periodicity of the 3CT (1s instead of 10s) and to avoid the connection issue with MQTT broker at start of HA.

Thx in advance.

harrymayr:
> Hello All,
> 
> Could you confirm if the PR #939 is inclued in the last official release (1.2.5) ? If not, it is planned for the next one ? I think it will be really appreciate, to win in reactivity with the periodicity of the 3CT (1s instead of 10s) and to avoid the connection issue with MQTT broker at start of HA.
> 
> Thx in advance.

No, it is not part of 1.2.5 and will not part of the next release.
There are plans to create a version which covers also other smart meters, but for now it's not clear, when this will be relased.

Issue #945, opened 2025-12-11
[Bug] strange behavior when Soc_min of one (ore more) device is reached - no P1 matching


CWILDMOUNTAIN:
### Home Assistant Version

2025.12.2

### Zendure Integration Version

(latest push of the repo + recent PR of harrymayr #939)

### Describe the bug

One (and shortly after my second of three) Hyper reached its SoC_min level. 
First thing I observed is, that the remaining device(s) are not matching the P1 anymore. 
Second thing: strangely one of the Hypers (in this very case Hyper 1) switches pass mode "on" and provides its solar power directly to home (of course it's not enough at this time of the year). The other empty Hyper No. 3 is not going into pass mode "on" but charges the empty battery (which is the expected behavior - I think).

My guess is: because one device is mistakenly in bypass, the smart matching can't be correct, as the manager expects 2 devices to supply power (which is not the case) 

(unfortunately I started the logs after Hyper 1 and 3 reached Soc empty, but you'll see Hyper 1 being in pass mode on and P1 is not matched) 

[home-assistant_zendure_ha_2025-12-11T13-15-57.427Z.log](https://github.com/user-attachments/files/24103460/home-assistant_zendure_ha_2025-12-11T13-15-57.427Z.log)

### To Reproduce

wait for one device to reach its Soc Min limit.

### Expected Behaviour

empty devices should not go into pass mode.
power distribution logic should only consider remaining devices. 

### What device are you using?

Hyper2000

### Log Extracts

```shell

```

PR #946, opened 2025-12-12
Handle empty devices on discharge

harrymayr:
Empty devices can deliver some power of the solarinput to the home-grid. But they could not be handled as "normal" devices which can discharge the full power. They should be taken out of the game and load the batteries with the amount of solarInput.

If they should deliver the solarpower to the grid, the maximum power available from the solarinput (- some offset [~30-40W] which can be seen on solarpower < 80W) need to be taken into account and not the full power the battery can deliver if it is not empty.

Should solve issue #945


CWILDMOUNTAIN:
thanks for the fix, will test it and report 👍

PR #994, opened 2026-01-06
Expose POWER_START / POWER_TOLERANCE as configurable Manager numbers (persisted + clamped)

oehrn:
## Summary
This PR exposes the existing smart-mode thresholds POWER_START and POWER_TOLERANCE as editable Home Assistant `number` entities on the **Zendure Manager** device.

## What changed
- Added two new `number` entities on the Zendure Manager device:
  - `number.zendure_manager_power_start` (POWER_START)
  - `number.zendure_manager_power_tolerance` (POWER_TOLERANCE)
- Values are **persisted via ConfigEntry options** so user settings survive Home Assistant restarts and are not reset to defaults.
- Implemented **dynamic constraints**:
  - POWER_START: min 50 W, max 5000 W
  - POWER_TOLERANCE: min 5 W, max = POWER_START - 10 W
  - If POWER_START is reduced below (tolerance + 10), tolerance is automatically clamped to the new maximum.
- Updated translations so the new entities are available in:
  - English, German, French, Dutch

## Testing / Release
- Changes have been successfully implemented and can be tested via the release **1.2.2** in this fork.

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -126,12 +126,15 @@ class ZendureDevice(EntityDevice):
         self.batteryInput = ZendureSensor(self, "outputPackPower", None, "W", "power", "measurement")
         self.batteryOutput = ZendureSensor(self, "packInputPower", None, "W", "power", "measurement")
         self.homeOutput = ZendureSensor(self, "outputHomePower", None, "W", "power", "measurement")
+        self.batInOut = ZendureSensor(self, "batInOut", None, "W", "power", "measurement", 0)
+        self.heatState = ZendureBinarySensor(self, "heatState")
         self.hemsState = ZendureBinarySensor(self, "hemsState")
         self.hemsStateUpdate = datetime.min
         self.availableKwh = ZendureSensor(self, "available_kwh", None, "kWh", "energy", None, 1)
         self.connectionStatus = ZendureSensor(self, "connectionStatus")
         self.connection: ZendureRestoreSelect
         self.remainingTime = ZendureSensor(self, "remainingTime", None, "h", "duration", "measurement")
+        self.nextCalibration = ZendureRestoreSensor(self, "nextCalibration", None, None, "timestamp", None)
 
         self.aggrCharge = ZendureRestoreSensor(self, "aggrChargeTotal", None, "kWh", "energy", "total_increasing", 2)
         self.aggrDischarge = ZendureRestoreSensor(self, "aggrDischargeTotal", None, "kWh", "energy", "total_increasing", 2)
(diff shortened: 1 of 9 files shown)


harrymayr:
Summary is fine, but it's not clear to me, what's your intention. I personally think to set the POWER_START perhaps to 30, but I have no glue, in whitch case I should set it to 5000. And to make it persistant, you easily can use ZendureRestoreNumber like the manual power is already defined.

oehrn:
What I’m trying to solve is pretty simple: the current defaults are “one-size-fits-all”, but in practice they aren’t. In my setup 50 W is too low, so smart mode starts reacting on tiny fluctuations and I’d rather push that threshold up. Making POWER_START / POWER_TOLERANCE editable lets people tune it to their install instead of being stuck with the defaults.

The 5000 W is just a guardrail, not a recommendation. I picked something high so it wouldn’t get in the way for bigger systems. Happy to change that to a more reasonable cap (e.g. match manual power limits / device constraints) if you think that’s better.

And yes: good call on ZendureRestoreNumber. That’s more consistent with manual power — I’ll update the PR this week to use that instead of ConfigEntry options.

FireSon:
@oehrn Thanks for the PR. There have been several request for more 'control' over the distribution mechanism, so you are not 'alone' if your change request! Still I do not think this approach is good for all users. I agree that it is a 'one size fits all', but what is the goal you want to achieve? What and why do you want to tune? And what is the gain? It is easy to change the constants if you want to improve the control. 
There has been a request to have an select entity where you can switch between an option which 'favours buying as little power as possible' and 'not export any solar energy'. These options would have different presets for certain distribution constants. That would be someting I would personally prefer, because it is easy to explain to all users.

At this moment I am refactoring the integration to make if fully 'local', so if I understand your underling reasons, I can take that into account. In that update it is for example possible to disable entities, which can help to hide 'advanced' entities by default.

On a side note, it would be good to reuse the formatting which is already used for the integration, it helps the readbility (and my programming ocd 😄)

Looking forward to your input!

oehrn:
Thanks for the feedback @FireSon — I get your point.

My goal is simply better, more predictable behavior across different installs. In my setup the defaults make smart mode react too early and “chase” small fluctuations. Being able to adjust the thresholds helps reduce that and makes the behavior feel more stable.

I agree that a user-friendly select like “minimize grid import” vs “avoid solar export” with presets is the better default for most users, because it’s easy to explain.

At the same time, I’d like to support both levels: simple presets for most users, and an optional “advanced” layer for people who want more control / transparency and the ability to fine-tune when needed. If your refactor allows hiding advanced entities by default, that sounds like a good fit.

oehrn:
Quick update reflecting the latest changes on my side.

Following @harrymayr’s suggestion, I’ve switched `power_start` and `power_tolerance` to `ZendureRestoreNumber` (same pattern as `manual_power`) and removed the ConfigEntry-options persistence. I also removed the hard upper clamp in the logic for `POWER_START` (HA still requires a max value for the UI, so the entity max is effectively “uncapped” via a very high guardrail).

Additionally, I ran into a startup/restore timing issue where Home Assistant could mark `number.zendure_manager_power_start` as “not provided” after restart. This was caused by restore callbacks trying to call `schedule_update_ha_state()` before the entities were fully attached, so I guarded those state updates to make restore/startup safe.

Re @FireSon: agreed that presets via a `select` might be the better default direction here. I’ll hold off changing the UI until there’s clearer agreement on what the presets should do.

If you prefer, I can just provide input and park this PR (on hold) or close it, and we can revisit with a preset-based approach later. I’m also fine iterating further (ranges/defaults/visibility) if that’s useful.

Kieft-C:
> There has been a request to have an select entity where you can switch between an option which 'favours buying as little power as possible' and 'not export any solar energy'. These options would have different presets for certain distribution constants. That would be someting I would personally prefer, because it is easy to explain to all users.

Totally agree. And I notice more and more user, which are jumping on the HA + Zendure Integration train, but they have no glue what it means and wonder why it is so "complicated" to set up.
I describe it often with, when a Apple child use first time windows.
I was a complete newbie to all coding and my learncurve is small, but what I see in all these forums and groups, show me, that the way allready it is, for the half of the user is overwhelming in a way.

So may then in general u need for all this new ideas a "expert" version or checking in setting up.
like with the local connection requesting at beginning. something, where u can give normal user a chance to get not overloaded with all selects.

FireSon:
@oehrn will integrate this into the new version, so I will close this PR for now. Will get back to you if it needs testing 😄

PR #1021, opened 2026-01-17
User defined optimal power & hems switch for hyper

harrymayr:


FireSon:
Thanks @harrymayr 
Since HEMS and HA are two different worlds and this can not be implemented for all devices, I will close this PR for now. I know that Zendure needs an call to their backend in order to manage HEMS properly, which makes it probably very difficult.

PR #1026, opened 2026-01-19
Fix invalid entity ID error in HA 2026.02

mithomas:
## Error

When using device names with spaces, for each entity there is an error like this:

```
Traceback (most recent call last):
File "/workspaces/ha-core/homeassistant/helpers/entity_platform.py", line 679, in _async_add_entities
await self._async_add_entity(
entity, False, entity_registry, config_subentry_id
)
File "/workspaces/ha-core/homeassistant/helpers/entity_platform.py", line 826, in _async_add_entity
raise HomeAssistantError(f"Invalid entity ID: {entity.entity_id}")
homeassistant.exceptions.HomeAssistantError: Invalid entity ID: switch.SolarFlow 800 Pro-lamp_switch
```

## Root Cause

Manually construction of entity ids worked until [HA PR #160302](https://github.com/home-assistant/core/pull/160302), which tightens entity id validation where it previously auto-corrected such ids.

## Fix

Removal of the manual entity_id assignment  works . since the entity already sets `_attr_has_entity_name = True`, leading Home Assistant to  automatically generate valid entity IDs with proper sanitization using the unique_id and device information.

### DIFF
--- a/custom_components/zendure_ha/entity.py
+++ b/custom_components/zendure_ha/entity.py
@@ -41,7 +41,6 @@ class EntityZendure(Entity):
             return
         self.device = device
         self._attr_unique_id = f"{self.device.name}-{uniqueid}"
-        self.entity_id = f"{entitytype}.{self.device.name}-{snakecase(uniqueid)}"
         self._attr_translation_key = snakecase(uniqueid)
         device.entities[uniqueid] = self
 


FireSon:
@mithomas I must say that this error is unknown to me. But by removing the entity_id, all existing entities will get different ids, which is for our current users not an option. We can consider it when we do a major release, because the entity_id's must be updated anyway. 
I will keep it in mind for that release! But I will close this PR for now.

harrymayr:
> @mithomas I must say that this error is unknown to me. But by removing the entity_id, all existing entities will get different ids, which is for our current users not an option. We can consider it when we do a major release, because the entity_id's must be updated anyway. I will keep it in mind for that release! But I will close this PR for now.

I don't think this is (totally) true, because the entities are already created. This PR apply only for new entities. I have it changed 5 days ago and did not run in any problems for now (restarted HA several times). The only problem I can imagine is, if you delete the integration and reinstall it again.

mithomas:
@FireSon:

>I must say that this error is unknown to me.

I encountered the issue with the the `dev`-branch of _HomeAssistant_ (which will soon be the 2026.2, I assume) - as indicated with the above link to the PR, it does not happen up to HA 2026.1.*

>But by removing the entity_id, all existing entities will get different ids, which is for our current users not an option.

This is not correct: Due to ` self._attr_unique_id = f"{self.device.name}-{uniqueid}"`, HA generates entity ids from the unique id which follow the same rule as the explicit `self.entity_id = f"{entitytype}.{self.device.name}-{snakecase(uniqueid)}"` (i.e. making it lowercase, replacing spaces...).

I confirmed this for my device _SF 800 Pro WZ-Balkon_ with clean HA installations:
* HA 2026.1.2 + `master` of this integration: `sf_800_pro_wz_balkon_available_energy`
* HA `dev` branch + this PR: `sf_800_pro_wz_balkon_available_energy`
 
So even for new entity creations or re-installations/-configurations, nothing changes.

FireSon:
Ahh that is very nice, I know when I started the integration, I had some trouble with it and that is why I added the line. In that case I will check it myself and reopen and approve the PR.
Thanks for the explanation, and learning something new 😉

PR #1029, opened 2026-01-19
Add single PV aggregation entities.

mithomas:
Include refactoring for device-specific PV input number.

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -141,6 +141,11 @@ class ZendureDevice(EntityDevice):
         self.aggrDischarge = ZendureRestoreSensor(self, "aggrDischargeTotal", None, "kWh", "energy", "total_increasing", 2)
         self.aggrHomeInput = ZendureRestoreSensor(self, "aggrGridInputPowerTotal", None, "kWh", "energy", "total_increasing", 2)
         self.aggrHomeOut = ZendureRestoreSensor(self, "aggrOutputHomeTotal", None, "kWh", "energy", "total_increasing", 2)
+
+        if self.solar_inputs > 1:
+            for i in range(1, self.solar_inputs + 1):
+                setattr(self, f"aggrSolar{i}", ZendureRestoreSensor(self, f"aggrSolarTotal{i}", None, "kWh", "energy", "total_increasing", 2,),
+            )
         self.aggrSolar = ZendureRestoreSensor(self, "aggrSolarTotal", None, "kWh", "energy", "total_increasing", 2)
         self.aggrSwitchCount = ZendureRestoreSensor(self, "switchCount", None, None, None, "total_increasing", 0)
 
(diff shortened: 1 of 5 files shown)


FireSon:
@mithomas Thanks for the input, much appreciated! 

I do not really know why we should add aggregate sensors for this. In the current version I think we already have very many sensors which are not used by most users. Also it is very easy to create an aggregate sensor yourself if it is really needed.  So what is the use case for it?
Secondly, at this moment, the sensors are created dynamically and it would have been much easier to add the aggregate sensor at the same time as the solarPower1 entity is created. Would be much less code 😄 

At this moment, we are reorganizing the integration, as a result the entities are not generated on the fly, but defined in code. In the same release it will be possible to make the entities disabled by default. This will help to limit the initial entities to the important ones. 

So, can you explain why these aggregate sensors are important to you? In any case I would like to wait to the next major release...

mithomas:
@FireSon: Thanks for your feedback!

>So what is the use case for it? [...] Also it is very easy to create an aggregate sensor yourself if it is really needed.

Mainly to track production differences between different panels over time. I currently integrate a _Hoymiles HM 800_ via _OpenDTU_ to MQTT and an _AP Systems EZ1-M_ via the _APsystems_ integration and both provide aggregated values per MPPT - thus I assumed that is standard and not something a users adds for himself:-).

Frankly I believe that this is something the device itself should do (support ticket open), but having it in this integration at least means that all aggregation sensors work identical.

>Secondly, at this moment, the sensors are created dynamically and it would have been much easier to add the aggregate sensor at the same time as the solarPower1 entity is created. Would be much less code 😄

You're absolutely right - thanks for the pointer! If you consider this PR, I'll amend this.

>In the same release it will be possible to make the entities disabled by default. This will help to limit the initial entities to the important ones.

That sounds like a good idea.

FireSon:
@mithomas You are right, it would be much better if the device would handle all the aggregation, let's see if they will implement it 😄 
Is there much difference between the OpenDTU entities and the current integration? Is it 'easy' to connect via OpenDTU , I have looked into it but it seemed a bit overwhelming at the time. I heard that the reaction times are much better is that correct? Are all Zendure devices supported?

Regarding the changes, I would propose that I will add the aggregate entities to the new version, and make them disabled by default. Is that OK?

mithomas:
@FireSon 

>Regarding the changes, I would propose that I will add the aggregate entities to the new version, and make them disabled by default. Is that OK?

Sounds perfect - thanks!

>Is there much difference between the OpenDTU entities and the current integration? Is it 'easy' to connect via OpenDTU , I have looked into it but it seemed a bit overwhelming at the time. I heard that the reaction times are much better is that correct? Are all Zendure devices supported?

_OpenDTU_ is actually a mapper from _Hoymiles HM/HMS_ devices to MQTT , so for _Zendure_ devices you won't need it as you can just use MQTT natively by setting it up in the app.

It is really easy to connect at least my SF 800 Pro via MQTT as _HomeAssistant_ auto-discovery is supported - meaning your device just shows up as an MQTT device with all the entities attached, no need to configure anything really. It supports most of the entities that the HTTP API has (except the ones for the fan atm - see https://github.com/mithomas/Zendure-HA/blob/feature/local-mqtt/mqtt_api_spec.yaml for a rough overview) and indeed the reaction times are better as you don't have to wait for a polling cycle (one a value changes, it's immediately pushed - if it doesn't, then there's silence).

As for devices I do not know - I can confirm the SF 800 Pro. I would assume that all devices that allow the setting of a custom MQTT server via app follow this schema.

By using the device via the MQTT integration one would have to implement smart matching for oneself, however - hence my feature request to support it here 😄: #1052

FireSon:
@mithomas I have added this to the new version, close this PR for now

Issue #1025, opened 2026-01-19
[Bug] Warning in the log for SolarFlow 800 Pro


m10z313:
### Home Assistant Version

2026.1.2

### Zendure Integration Version

v1.2.4

### Describe the bug

I've owned the SolarFlow 800 Pro for a few days now and immediately installed the Home Assistant integration. After updating to version 1.2.4 and restarting Home Assistant, I noticed the following warning in the logs:

Logger: homeassistant.components.sensor.recorder
Quelle: components/sensor/recorder.py:439
Integration: Sensor (Dokumentation, Probleme)
Erstmals aufgetreten: 17. Januar 2026 um 11:30:10 (1 Vorkommnis)
Zuletzt protokolliert: 17. Januar 2026 um 11:30:10

Entity sensor.solarflow_800_pro_aggr_grid_off_power_total from integration zendure_ha has state class total_increasing, but its state is negative. Triggered by state -1.4166337447279 with last_updated set to 2026-01-17T10:24:59.999999+00:00. Please create a bug report at https://github.com/zendure/zendure-ha/issues

I should mention that I currently have the inverter of my balcony solar system connected to the off-grid outlet. Therefore, I suspect that the sensor mentioned in the warning is currently configured for power output only, and not power input.

### To Reproduce

1. Update Integration to Version v1.2.4
2. Restart Home Assistant
3. Open the logs

### Expected Behaviour

No warning should be displayed.

### What device are you using?

SF800 Pro

### Log Extracts

```shell
Logger: homeassistant.components.sensor.recorder
Quelle: components/sensor/recorder.py:439
Integration: Sensor (Dokumentation, Probleme)
Erstmals aufgetreten: 17. Januar 2026 um 11:30:10 (1 Vorkommnis)
Zuletzt protokolliert: 17. Januar 2026 um 11:30:10

Entity sensor.solarflow_800_pro_aggr_grid_off_power_total from integration zendure_ha has state class total_increasing, but its state is negative. Triggered by state -1.4166337447279 with last_updated set to 2026-01-17T10:24:59.999999+00:00. Please create a bug report at https://github.com/zendure/zendure-ha/issues
```

PR #1031, opened 2026-01-19
Fix aggr sensors for gridoff

harrymayr:
Fixes issue #1025

### DIFF
--- a/custom_components/zendure_ha/devices/solarflow800Pro.py
+++ b/custom_components/zendure_ha/devices/solarflow800Pro.py
@@ -18,7 +18,7 @@ class SolarFlow800Pro(ZendureZenSdk):
         self.setLimits(-1000, 800)
         self.maxSolar = -1200
         self.offGrid = ZendureSensor(self, "gridOffPower", None, "W", "power", "measurement")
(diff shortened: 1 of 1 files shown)


custom_components/zendure_ha/devices/solarflow2400ac.py
```diff
@@ -18,7 +18,7 @@ def __init__(self, hass: HomeAssistant, deviceId: str, prodName: str, definition
         self.setLimits(-2400, 2400)
         self.maxSolar = -2400
         self.offGrid = ZendureSensor(self, "gridOffPower", None, "W", "power", "measurement")
-        self.aggrOffGrid = ZendureRestoreSensor(self, "aggrGridOffPowerTotal", None, "kWh", "energy", "total_increasing", 2)
+        self.aggrOffGrid = ZendureRestoreSensor(self, "aggrGridOffPowerTotal", None, "kWh", "energy", "total", 2)
```
FireSon: Why should this be different than all others? Just curious...

harrymayr: Could be positive **and** negative. Depending how you use it. If you connect an inverter, it isn't _increasing_ anymore 😉

Issue #1053, opened 2026-01-28
[Bug] SF2400 AC smart mode stays to 0 and won't discharge event if zendure manager is in smart mode


fabhacs:
### Home Assistant Version

2026.1.3

### Zendure Integration Version

v1.2.4

### Describe the bug

Hi, I’m experiencing the same kind of issue as https://github.com/Zendure/Zendure-HA/issues/964

What I can see on my side, is that the zendure manager is in smart mode, but the Solarflow 2400 AC smart mode stays on 0 

-> Important note : It used to work perfectly in the past
This issue is recent, but I don’t know which firmware/software version introduced it.

### To Reproduce

Repro steps / 2 examples:

1. Charge in Manual mode → switch to Smart mode: the SF 2400 remains “Idle” and won’t discharge.

2. Smart mode + Shelly as smart meter: during rapid transitions between charge/discharge, the SF 2400 intermittently falls back to “Idle”.

**Workaround**: setting an output limit briefly “unlocks” it, after which it resumes normal adaptive discharging.

### Expected Behaviour

When zendure manager is in smart mode, the SF 2400 should never goes to "idle" mode (except if the SOC is at its maximum or minimum obviously)

### What device are you using?

SF2400 AC
Device controlled on Zendure Integration via zenSDK
HEMS **disabled** on zendure app

### Log Extracts

```shell

```

harrymayr:
Would be nice, if you can provide a debug log.

fabhacs:
I don't know where I can find this kind of log.
I have no errors in HA, I just see that the battery is not discharching even though Zendure Manager is on Smart Mode.

harrymayr:
You can checkout the [wiki](https://github.com/Zendure/Zendure-HA/wiki/Function-description) with respect to Function description and setup. There are also some tutorials how to create a log file and other things.

fabhacs:
As I said it worked perfectly until recently.
I will check if I can find some logs when this happens

harrymayr:
Perhaps because every change need to be triggered by a P1 sensor, which show the need/surplus of the home grid. If there is no sensor which update i.e. every 5 seconds, than the devices also will not get an update.
This may be a difference to older versions.

fabhacs:
I think this problem were introduced with v1.2.0
I noticed that from then, I needed to force the output limit so that the SF "wake up" and actually do the smart stuff it was supposed to do

fabhacs:
[...]

PR #1055, opened 2026-01-28
Modify power discharge handling for charging devices

harrymayr:
Adjust power discharge logic for devices to handle grid input power correctly.


fabhacs:
Fixes #1053

PR #1065, opened 2026-02-03
User defined optimal power & hems switch for hyper

HenryMoec:


FireSon:
@HenryMoec Thanks for the PR! 
There have been several PR's making the distribution settings 'editable' which I did not merge. The reason for that is to keep it as simple as possible. In the next new version, the distribution changes slightly, and we will add a some disabled settings for the experienced users. So please bear with us, and give your input/thoughts on how we can improve that once the new major version is out. 
The HEMS switch, is more a design principle. The next major release is aimed to be cloud independent, but we will still be able to use cloud connection. In combination with HEMS this will increasingly be more difficult to manage., because switching to HEMS in local mode does not make sense. Zendure is looking into the possibility to add the switch functionality to other devices as well, but indicated that it is 'not that simple'. On the other hand, I do not understand why HEMS would make sense if you use HA. Is it performing better? Or is there another reason to use it?

So I will close the PR for now, but if you still think it is useful, let me know and we can discuss possibilities 😄

HenryMoec:
@FireSon 
Many thanks for your comments. At first, the HEMS switch wasn‘t a request from my side. It was merged from an other issue to this PR. I don‘t need such a switch, as, as you said, it didn‘t make sense, if HA is used.
My request was the power distribution issue.
Currently I‘m using patched v1.1.3, as the 85% distribution is simple and efficient. In my opinion, the power distribution in later version is a little bit over engineered.
But lets see, how the new release works, related to the distribution. I will let you know, how it works for me.
For me, the most important point in the power distribution, is if the devices works in an efficient area. As you manage several devices in Z-HA with different power areas, in my thought an adjustable Distribution is the only way to make all users happy.
Greatings 
Henry

PR #1100, opened 2026-02-14
Update version to 1.2.5 in manifest.json

Kieft-C:
should always show latest version plus optional small infos - why 1.0.0 back?

### DIFF
--- a/custom_components/zendure_ha/manifest.json
+++ b/custom_components/zendure_ha/manifest.json
@@ -17,5 +17,5 @@
     "stringcase==1.2.0"
   ],
   "single_config_entry": true,
-  "version": "1.0.0"
+  "version": "1.2.5"
 }


FireSon:
Hi @Kieft-C the version is set by the release workflow, which is very convenient. That is the reason I changed it back to 1.0.0 so it is clear. Hope that explains it 😄

Kieft-C:
Hi @FireSon

Only important for "inbetween" CodeZip Downloader is then, they have no fast check where they are.
They not know the last version and see only 1.0.0 when manual installed the integration.

FireSon:
Good point, but is is very difficult to update it at every change, if it is 1.0.0 you are certain that you use a 'custom' version.

cybe:
It might be confusing when the content of release artifacts differ from the tagged commit it is based on.

Wouldn't the workflow be, to commit a version update such as this change. Tag it, then create a Github release? I think that is how most other HACS integrations implement this. Other than semantic release bots and whatnot.

FireSon:
@cybe if you have an example of such an approach, please let me know, When I started the integration, this was the only example I could find, maybe there are new possibilities 😄

PR #1114, opened 2026-02-20
Fix discharge state display when min SOC limit is reached

eremid:
## Problem

The operation state was incorrectly set to `DISCHARGE` based solely on
`setpoint > 0` (house power demand), without checking whether any device
was actually able to discharge.

When all batteries reached their min SOC limit, devices were moved to the
idle list (`homeOutput == 0`), but the manager kept displaying
**"Discharging"** instead of **"Standby"**.

## Fix

Only set `DISCHARGE` state when `self.discharge` is non-empty, meaning at
least one device is actively outputting power to the home.

```python
# Before
ManagerState.DISCHARGE.value if setpoint > 0 else ManagerState.IDLE.value

# After
ManagerState.DISCHARGE.value if setpoint > 0 and self.discharge else ManagerState.IDLE.value
```

## Impact

- No functional change to power distribution logic
- Corrects the displayed operation state in Home Assistant when min SOC is reached

### DIFF
--- a/tests/test_manager.py
+++ b/tests/test_manager.py
@@ -1,0 +1,93 @@
+"""Unit tests for ZendureManager.power_discharge() operation state logic.
+
+Covers the three combinations of setpoint / discharge-list that determine
+whether the manager reports DISCHARGE or IDLE.
+"""
+
+from __future__ import annotations
+
+import sys
+from datetime import datetime
+from pathlib import Path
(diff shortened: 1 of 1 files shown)


FireSon:
@eremid Thanks again! it is a bit confusing for people. As said in my other comment, I will keep out the Unittest for now.

PR #1116, opened 2026-02-21
Fix entity write sending snake_case property names to device

eremid:
## Problem

`entityWrite()` was building the property name by stripping the device name prefix from `entity.unique_id`. Since `unique_id` is converted to snake_case at creation time (via `snakecase()`), the property name sent to the device was also in snake_case.

For example:
- `lampSwitch` → sent as `lamp_switch`
- `minSoc` → sent as `min_soc`
- `socSet` → sent as `soc_set`
- `buzzerSwitch` → sent as `buzzer_switch`

The device silently ignores unknown property names, so all write commands from Home Assistant had no effect, while the Zendure app (which sends the correct camelCase names) worked fine.

This affected **all writable entities** (switches, selects, numbers) on all device types using both `ZendureDevice.entityWrite` (MQTT) and `ZendureZenSdk.entityWrite` (HTTP local API).

## Fix

Store the original camelCase key as `property_key` on `EntityZendure` at initialisation time, and use it directly in both `entityWrite()` implementations instead of reconstructing it from the snake_case `unique_id`.

## Changes

- `entity.py`: add `self.property_key = uniqueid` before the snake_case conversion
- `device.py`: replace `entity.unique_id[(len(self.name) + 1):]` with `entity.property_key` in `ZendureDevice.entityWrite` and `ZendureZenSdk.entityWrite`

## Testing

Tested successfully on **Hyper 2000** and **SolarFlow 2400 AC+** in local mode:
- LED switch (`lampSwitch`) now turns on/off correctly from Home Assistant
- Minimum SOC (`minSoc`) and SOC target (`socSet`) now apply correctly
- Buzzer switch (`buzzerSwitch`) confirmed working

### DIFF
--- a/tests/test_entity_write.py
+++ b/tests/test_entity_write.py
@@ -1,0 +1,103 @@
+"""Unit tests for EntityZendure.property_key and ZendureDevice.entityWrite().
+
+Covers two aspects of the same fix:
+  1. EntityZendure.__init__ stores the original camelCase key as property_key
+     before it is converted to snake_case for unique_id.
+  2. entityWrite() uses property_key (camelCase) in the MQTT payload,
+     not a name derived from unique_id (which is snake_case).
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from pathlib import Path
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+REPO_ROOT = Path(__file__).resolve().parents[1]
+sys.path.insert(0, str(REPO_ROOT))
+
+# conftest stubs are already in sys.modules when pytest loads this file.
+from custom_components.zendure_ha.device import ZendureDevice  # noqa: E402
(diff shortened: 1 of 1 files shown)


FireSon:
@eremid Thanks for noticing this! I did miss it 😄 

Since HA2026.2 there has been a problem with the entity_id's and this bug is a result of that. In the next release we will modify all entity_id's in order to align them for the new development of a 'local only' version of the integration. I have fixed this in a different way, by using the translation_key, and removing all '_' characters. The Zendure devices are not case sensitive to properties, so no 'new' property is necessary.

At this moment we do not have unit tests in the integration. To be honest, I do not know if it is really helpful at this moment, but I have planned to look into that at another time. This integration was my first Python experience, so I still need to learn a lot 😉 

For now, I will close the PR, but if you think I have missed something, please let e know!

Issue #1127, opened 2026-02-23
[Bug/Missing Feature] Migration of Template Sensors missing for 1.2.6


deveth0:
### Home Assistant Version

2026.2.3

### Zendure Integration Version

v1.2.6-pre1

### Describe the bug

Migration of all sensors, utility meters etc to the new naming schema worked fine but the template sensors have not been touched.
It would be nice, if those could be migrated automatically too. 

### To Reproduce

Add Zendure Device with 1.2.5, create Template sensor using one of it's entities and install 1.2.6

### Expected Behaviour

Template sensor is modified to reflect the new name change

### What device are you using?

SF800

### Log Extracts

```shell

```

PR #1131, opened 2026-02-23
Implement template patching for device renaming

harrymayr:
Added methods to patch template helpers and update device entities.


harrymayr:
This should fix issue #1127

PR #1133, opened 2026-02-24
Fix renaming of helper sensors

harrymayr:


harrymayr:
Still does not work with Entries like this : "options":{"entity_ids":["sensor.fo2nen8n3401037_hochste_zelltemperatur"],...
😒

FireSon (approved):

PR #1141, opened 2026-03-01
feat: add max_battery_output_power sensor to ZendureManager

eremid:
This pull request introduces a new `max_battery_output_power` sensor to the Zendure integration.

### Need
The purpose of this sensor is to provide the actual total discharge capacity of the setup, considering both hardware limits and fuse group constraints. This allows users to:
- **Calculate battery utilization percentage:** By comparing current power output to this maximum value, users can create informative dashboards or widgets showing how much of their discharge capacity is being used.
- **Accurate monitoring:** The sensor dynamically updates based on the current configuration, including when devices are added or removed from fuse groups.

### Key Changes
- **New Sensor:** Added `max_battery_output_power` to `ZendureManager`.
- **Intelligent Calculation:**
  - Sums the hardware discharge limits of all connected devices with batteries (`kWh > 0`).
  - Caps the sum for each fuse group by its `maxpower` configuration.
  - Aggregates the results across all fuse groups.
- **Dynamic Updates:** The sensor value is refreshed when fuse groups change or during regular manager updates.
- **Translations:** Added support for English, German, French, and Dutch.

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -101,6 +101,7 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         self.manualpower = ZendureRestoreNumber(self, "manual_power", None, None, "W", "power", 12000, -12000, NumberMode.BOX, True)
         self.availableKwh = ZendureSensor(self, "available_kwh", None, "kWh", "energy", None, 1)
         self.power = ZendureSensor(self, "power", None, "W", "power", "measurement", 0)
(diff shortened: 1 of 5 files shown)


FireSon:
Hi @eremid I understand the idea, but I am a bit hesitant to add too many entities, since it becomes more difficult for new users to understand. It is clear you cannot access these values yourself. The place where you added the update is called very often, which to keep everything as smooth as possible is not ideal. In another PR I added something similar, the totalKwh. Let me think about fixing them together!

PR #1162, opened 2026-03-06
fix: prevent setpoint sign flip for SOCFULL devices (issue #1151)

### LINKED ISSUE #1151: [Bug] Smart power distribution mode not working once all batteries are fully charged and bypass forbidden
### Home Assistant Version

2026.2.3

V. 1.2.5

### Describe the bug

2xHyper 2000, both with 3x AB2000X, Bypass forbidden, because I don't want to chase >2kW senselessly into the grid and burn up the inverters in the process.

Once both Hyper reaches 100% SoC the output cycles constant between zero and the needed output for the home grid.
Staying at zero for minutes, then step up to 45 - 50W, rising slowly to the matching power and then drops to zero after only a few seconds and so on until solar input is below needed power and discharging starts.

### To Reproduce

Use inverter and batterie at 100% SoC as described with solar surplus. 

### Expected Behaviour

Distributing matching power to the home grid.

Hyper2000

### Log Extracts

```shell

```

### Other Information

_No response_

swissky:
## Problem

Fixes #1151 — *Smart power distribution mode not working once all batteries are fully charged and bypass forbidden*

### Root Cause

In `powerChanged()`, when a device is at `SOCFULL` (100% SoC), its solar production is accumulated into `discharge_bypass`:

```python
# manager.py line ~425
self.discharge_bypass -= d.pwr_produced if d.state == DeviceState.SOCFULL else 0
```

(`pwr_produced` is negative = solar excess, so `discharge_bypass` grows positive)

Then later:

```python
# manager.py line ~443
if self.discharge_bypass > setpoint:
    setpoint -= self.discharge_bypass
```

**The bug**: when `discharge_bypass > setpoint`, the subtraction flips `setpoint` from positive (home needs power) to negative — even when the original `p1` reading was `>= 0`. This incorrectly triggers `power_charge()` instead of `power_discharge()`, causing the output to drop to 0W.

The effect is most visible when:
- All batteries are at 100% SoC
- Bypass is set to **Forbidden**
- Smart/smart_discharging/store_solar mode is active
- Solar fluctuates slightly (making `pwr_produced` near-zero intermittently)

The output then cycles: 0W → slowly ramps up → drops to 0W again, repeating for minutes.

### Fix

Clamp `setpoint` so it never drops below `0` when the original `p1` value was non-negative (i.e. the home is consuming power, not exporting). When `p1 < 0` (genuine grid surplus), the full subtraction is still applied to avoid unintended export.

```python
if self.discharge_bypass > 0:
    setpoint = max(0 if p1 >= 0 else setpoint - self.discharge_bypass,
                   setpoint - self.discharge_bypass)
```

### Community-confirmed workarounds (until this fix is merged)
1. Switch to `manual` mode with a fixed output limit
2. Use `store_solar` mode (reduces cycling but doesn't fully eliminate it)  
3. Enable CT mode on one device via the Zendure app (reported by @Zen4712 in #1151)

### Tested with
- 2× Hyper 2000 + AB2000X, bypass forbidden, SoC = 100%
- Simulated via `simulation.csv` output — setpoint stays positive after fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <p align="center">
-  <img src="https://zendure.com/cdn/shop/files/zendure-logo-infinity-charge_240x.png?v=1717728038" alt="Logo">
+  <img src="https://zendure.com/cdn/shop/files/logo.svg" alt="Logo">
 </p>
 
 # Zendure Home Assistant Integration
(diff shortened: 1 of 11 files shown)


Herr-Frodo:
That seems not to be the solution - AC output stucks at value of POWER_START from constant.py per device.

FireSon (approved):

Issue #1156, opened 2026-03-05
[Bug] Controls aren't working anymore with Hub2000 since 1.2.6


punzenbergerpeter:
### Home Assistant Version

2026.3.0

### Zendure Integration Version

v1.2.6

### Describe the bug

After updating to version 1.2.6, none of the controls are functioning anymore. It appears that the control settings are no longer being transmitted to Zendure, preventing any changes from being applied.

However, settings modified within the Zendure app are successfully sent to Home Assistant and are displayed correctly.

### To Reproduce

Go to Hub2000-Device in HomeAssistant
Change any setting (for example Output Limit)
Go to Zendure App
Output Limit is not changed
-----------------------------------
Change Output Limit in Zendure App
Look at Output Limit in Home Assistant
Output Limit is changed to setting from Zendure App

### Expected Behaviour

I expect that changing settings in HomeAssistant leads to changes at Zendure.

### What device are you using?

Hub2000

### Log Extracts

```shell
nothing in the system logs
```

Syltin:
Confirmed. Same behavior since 1.2.6 on hyper 2000

Hamsterberg:
I can confirm that too. 

Version: 1.2.6-pre2
2x AIO2400
1x Solarflow800 pro
Zendure app 6.0.1

With HA integration, you can only monitor at the moment; control is only possible via the Zendure app.

AndreR84:
Same behavior on my HUB 2000 since 1.2.6

therealchebo:
I also cant change the mininum SoC of the batteries in HA (for exapmple to 40 %). in the app it works.
i use a SF1200 with 3 x AB1000

Syltin:
Looks like 1.2.7 pre3 made controls from ha to Zendure via cloud back to expected behavior

PR #1172, opened 2026-03-09
fix the device not responding to changes of the acMode after renaming the ids

zaubara:
I wasn't able to charge my Hyper in the new release by using "off" mode, and the device didn't seem to respond to changes to the acMode.
Using negative values on manual mode made the device charge, but the acMode inexplicably reverted to output (after manually changing them to input).
This should fix both issues.


punzenbergerpeter:
This PR fixes #1156 

Thanks for the fix.

FireSon:
Thanks @zaubara several other have solved the same problem as well. But many thanks for the time, much appreciated. I went with the other solution, since it did not require a separate variable.

Issue #1084, opened 2026-02-09
[Feature] Adjustable Power Distribution requested


HenryMoec:
### Describe the feature

Dear developers,

I‘m using Z-HA since a couple of month and I always was really happy with it. Really a big thank to you guys for the fantastic Integration.
But since v1.2.x I lthe power distribution in case of more than 1 device, is not efficient. The splitting of the power is done to early. Eg sometimes Z-HA splits 400w into 2x200w.
Ive made an analysis over a couple of weeks for my SF2400 to see how the efficiency looks
over the complete power range. Attached you can see the results. As you can see, the SF2400 needs at least approx 600-800w to work in an efficient area, so splitting should not be done before reaching approx 1400w cummulated power.
So I unfortunatly cannot use Z-HA in the newest version. Currently Im using 1.1.3 with patch from PR1026.
I already opened an issue related to this issue, where Harry created a PR, but this PR was closed because there also was a kind of HEMS switch merged in it.
So I really would appreciate, if in further versions an adjustable power distribution gate can be integrated.
Many thanks in advance.

Henry

<img width="1189" height="640" alt="Image" src="https://github.com/user-attachments/assets/64b5a91c-0593-48ff-8a50-b426febcdf02" />

### What device is this for?

SF2400 AC

FireSon:
@HenryMoec yes that is something we will improve in the next version. Thanks for the feedback!

HenryMoec:
If someone will make such an efficiency measurement for other devices, codes are provided here:
[https://github.com/HenryMoec/Wirkungsgradermittlung-Zendure-Ger-te](https://github.com/HenryMoec/Wirkungsgradermittlung-Zendure-Ger-te)
Repo isn`t finished yet, but the codes are provided.

PR #1194, opened 2026-03-15
feat: weight power distribution by battery capacity (kWh)

Nicolas-nwb:
## Problem

When mixing batteries of different capacities (e.g. SolarFlow 2400AC with 17 kWh and Hyper 2000 with 4 kWh), the power distribution is unfair. The current weight formula `pwr_max * SOC` ignores total energy capacity, so a 17 kWh battery only receives ~56% of discharge power despite holding ~73% of the total stored energy.

This has been reported in #806 and #1084.

## Solution

Multiply the distribution weight by `max(d.kWh, 1.0)` so that larger batteries receive proportionally more charge/discharge power.

**New formula:** `weight = pwr_max × SOC × max(kWh, 1.0)`

6 lines changed in `manager.py`:
- `powerChanged()` — weight accumulation for charge and discharge
- `power_charge()` — per-device power calculation and weight decrement
- `power_discharge()` — per-device power calculation and weight decrement

## Backward compatibility

| Scenario | Behavior |
|---|---|
| Same-capacity batteries | kWh factor cancels out → **no change** |
| Single battery | Weight ratio is always 100% → **no change** |
| kWh not yet known (0.0) | Fallback `max(0.0, 1.0) = 1.0` → **original behavior** |
| Physical power limits | Still enforced via `min(pwr, d.pwr_max)` / `max(pwr, d.pwr_max)` caps |

## Test results

Tested on a real setup with 3 batteries (SolarFlow 2400AC 17 kWh + 2× Hyper 2000 4 kWh):

| Battery | Before | After |
|---|---|---|
| SolarFlow 2400AC (17 kWh, 71% SOC) | 0.89 kW (56%) | 1.56 kW (93%) |
| Hyper 2000 (4 kWh, 56% SOC) | 0.35 kW (22%) | 0.12 kW (7%) |
| Hyper 2000 2 (4 kWh, 55% SOC) | 0.34 kW (22%) | idle |

The larger battery now carries the bulk of the discharge, which is the expected behavior given its much higher stored energy. Peak power limits (2400W / 1200W) are still respected.

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -421,7 +421,7 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                     self.charge.append(d)
                     self.charge_limit += d.fuseGrp.charge_limit(d)
                     self.charge_optimal += d.charge_optimal
-                    self.charge_weight += d.pwr_max * (100 - d.electricLevel.asInt)
+                    self.charge_weight += d.pwr_max * (100 - d.electricLevel.asInt) * max(d.kWh, 1.0)
                     setpoint += home
                 # SOCEMPTY means, it could not discharge the battery, but it is still possible to feed into the home using solarpower or offGrid
                 elif (home := d.homeOutput.asInt) > 0:
@@ -430,7 +430,7 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                     self.discharge_limit += d.fuseGrp.discharge_limit(d)
                     self.discharge_optimal += d.discharge_optimal
                     self.discharge_produced -= d.pwr_produced
-                    self.discharge_weight += d.pwr_max * d.electricLevel.asInt
+                    self.discharge_weight += d.pwr_max * d.electricLevel.asInt * max(d.kWh, 1.0)
                     setpoint += home
 
                 else:
@@ -510,8 +510,8 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         limit = self.charge_limit
         setpoint = max(limit, setpoint)
         for i, d in enumerate(sorted(self.charge, key=lambda d: d.electricLevel.asInt, reverse=True)):
-            pwr = int(setpoint * (d.pwr_max * (100 - d.electricLevel.asInt)) / self.charge_weight)
-            self.charge_weight -= d.pwr_max * (100 - d.electricLevel.asInt)
+            pwr = int(setpoint * (d.pwr_max * (100 - d.electricLevel.asInt) * max(d.kWh, 1.0)) / self.charge_weight)
+            self.charge_weight -= d.pwr_max * (100 - d.electricLevel.asInt) * max(d.kWh, 1.0)
 
             # adjust the limit, make sure we have 'enough' power to charge
             limit -= d.pwr_max
@@ -561,9 +561,9 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         setpoint = min(limit, setpoint)
         for i, d in enumerate(sorted(self.discharge, key=lambda d: d.electricLevel.asInt, reverse=False)):
             # calculate power to discharge
-            if (pwr := int(setpoint * (d.pwr_max * d.electricLevel.asInt) / self.discharge_weight)) < -d.pwr_produced and d.state == DeviceState.SOCFULL:
+            if (pwr := int(setpoint * (d.pwr_max * d.electricLevel.asInt * max(d.kWh, 1.0)) / self.discharge_weight)) < -d.pwr_produced and d.state == DeviceState.SOCFULL:
                 pwr = -d.pwr_produced
-            self.discharge_weight -= d.pwr_max * d.electricLevel.asInt
+            self.discharge_weight -= d.pwr_max * d.electricLevel.asInt * max(d.kWh, 1.0)
 
             # adjust the limit, make sure we have 'enough' power to discharge
             limit -= -d.pwr_produced if solaronly else d.pwr_max


FireSon:
I did not test this is real life, but the goal of the current software, is that all batteries are discharged at more or less the same electriclevel. Is that not the case? Would in your case not 1 device to all the work? I am just trying to understand, I do not have the same configuration as you...

Nicolas-nwb:
Hi FireSon, thanks for taking the time to review this PR! This is my first contribution to the project, so I appreciate the feedback.

You're right that the current software tries to discharge all batteries at the same SOC level. The issue is that with batteries of very different capacities, **equal SOC does not mean equal energy contribution**.

Here's a concrete example from my setup:

| Battery | Capacity | SOC | Stored energy |
|---|---|---|---|
| SolarFlow 2400AC | 17 kWh | 70% | 11.9 kWh |
| Hyper 2000 | 4 kWh | 70% | 2.8 kWh |

Both are at 70% SOC, but the 2400AC holds **4.25x more energy**. Without the kWh factor, the weight formula `pwr_max × SOC` gives them similar power shares, so the small battery empties much faster than the large one. They end up at different SOC levels despite the software trying to keep them balanced.

With the kWh factor, the 17 kWh battery naturally receives more power, which means **both batteries actually reach the same SOC at the same time**. The kWh weighting doesn't fight the SOC balancing goal; it makes it work correctly for mixed-capacity setups.

To address your concern about one device doing all the work: no, because the SOC term is still in the formula. As a battery discharges and its SOC drops, its weight decreases, so the other batteries progressively take over. Both batteries still participate and **absorb peaks together**: if there's a large charge or discharge spike, the power is split across all active batteries proportionally to their capacity, which also reduces stress on individual devices.

For setups where all batteries have the same capacity, the kWh factor cancels out (same multiplier on both sides), so it changes absolutely nothing.

**Possible improvement:** the current formula uses total `kWh`, but doesn't account for `minSoc`. A battery with 17 kWh and minSoc 10% has 15.3 kWh of usable capacity, while a 4 kWh battery with minSoc 30% only has 2.8 kWh usable. The code already computes `availableKwh = (electricLevel - minSoc) / 100 × kWh` per device. A further refinement could use effective capacity `kWh × (100 - minSoc) / 100` instead of raw `kWh` in the weight formula. I'll experiment with this on my setup and can submit a follow-up PR if the results are good.

harrymayr:
I just checked the actual implementation and created a EXCEL-Sheet for this. But in my opinion, the capacity is implizit already included. All batteries are empty at nearly the same time. So I don't think there is a need for changing that. (I did not include the checking on optimal_power, which will change only slightly the result).
image
[Discharge_Distribution.xlsx](https://github.com/user-attachments/files/26347774/Discharge_Distribution.xlsx)

Nicolas-nwb:
Hi @harrymayr Thanks, 

Your math is correct: without minSoc, all batteries do reach 0% at roughly the same time.

However, in practice, batteries stop discharging when they hit `minSoc`. I took your spreadsheet and added two columns — **"Effective total (minSoc ≥ B6%)"** and **"Grid import"** — plus a configurable `minSoc` parameter in cell B6. Nothing else was changed; your formulas and values are untouched.

image

[Discharge_Distribution_with_minSoc.xlsx](https://github.com/user-attachments/files/26349837/Discharge_Distribution_with_minSoc.xlsx)

At 13:00, both Hypers cross below 10% and go `SOCEMPTY`, while the SF is still at 31%. From that point, only the SF contributes (capped at 2400W), leaving a **1100W deficit from the grid for ~1h40**.

The theoretical convergence to 0% is real, but `minSoc` prevents it from happening in practice. The small batteries exhaust their usable range much earlier, and the system loses peak capacity. Both batteries still participate and absorb peaks together, but only as long as they're above minSoc.

With the kWh factor, the SF carries more of the load (proportional to its 17 kWh vs 4 kWh), so all batteries cross 10% within ~40 minutes of each other instead of 1h40. The house stays closer to 3500W for longer.

You can try it yourself: I attached your spreadsheet with just the two extra columns added. Change B6 to any minSoc value and see the impact.

For setups where all batteries have the same capacity, the kWh factor cancels out (same multiplier on both sides), so it changes absolutely nothing.

**Possible improvement:** the current formula uses total `kWh`, but a further refinement could use effective capacity `kWh × (100 - minSoc) / 100` instead of raw `kWh`, since the code already computes `availableKwh` per device. I'll experiment with this on my setup and can submit a follow-up PR if the results are good.

harrymayr:
Both EXCEL files are not "correct". I added now the limiting at the discharge_limits and the starting of devices according to the optimal_power.
Finaly it is depending on the power which is demanded, if there is an advantage to increase the power for the SF2400 or not. At some point the discharge-limit is the limiting factor.
The EXCEL formulas become a little bit confused , so I don't want to add more features.
In general I aggree to use a "relative" SoC related to the available capacity. But this is only necessary for higher power demands for longer times, which need 2-3 devices to discharge. For demanded power which can be fulfilled from one device, there is no issue.
[Discharge_Distribution_with_minSoc_corrected.xlsx](https://github.com/user-attachments/files/26359405/Discharge_Distribution_with_minSoc_corrected.xlsx)

Nicolas-nwb:
You're right, at lower demand where one device can handle it alone, the kWh factor doesn't change anything. The benefit only shows up when multiple devices discharge together at higher demand, which is my use case. And for setups where all batteries have the same capacity, the kWh multiplier is identical on both sides, so it cancels out completely: zero behavioral change.

That said, I think mixed-capacity setups will become more and more common as the project grows. I started with a Hyper 2000 myself and later added a SolarFlow 2400AC. It's a natural upgrade path, and this change would make the transition smoother for anyone doing the same.

I'm happy to keep this on my fork if you'd rather not merge it. Either way, thanks for taking the time to analyze it in detail. I came here to improve power distribution and ended up leveling up my Excel skills too!

Issue #1015, opened 2026-01-16
[Bug] setpoint incorrect while consuming power from GridOff Socket


fellek:
### Home Assistant Version

2026.1.1

### Zendure Integration Version

v1.2.2

### Describe the bug

Solar: Connected to AC-Grid
Device: SolarFlow 2400 AC
GridOffPower active

Zendure-HA integration has difficulties with zero feed-in while GridOff power is being consumed. As long as solar power is available and GridOff Socket is used, approximately the negative value of power consumed by GridOff Socket will be permanently fed into the public grid.

<img width="781" height="388" alt="Image" src="https://github.com/user-attachments/assets/c8b664f6-9e8d-412e-91a0-11ae05a23a3e" />

### To Reproduce

Turn On Smartcharge
Off-Grid-Normal
Consume Power from GridOff-Socket (60-90 W)
Wait for SolarPower feed-in 400-500W

Unplug GridOff devices

Solarflow starts charging with zero feed-in.

### Expected Behaviour

Consuming GridOff power should not disturb zero feed-in. 

### What device are you using?

SF2400 AC

### Log Extracts

```shell
2026-01-16 12:01:34.580 INFO (paho-mqtt-client) [custom_components.zendure_ha.api] Topic: /***/SolarFlow 2400 AC/time-sync => {'messageId': 226179}
2026-01-16 12:01:34.680 INFO (paho-mqtt-client) [custom_components.zendure_ha.api] Topic: iot/***/SolarFlow 2400 AC/time-sync/reply => {'zoneOffset': '+01:00', 'messageId': 465705, 'userId': ***, 'timestamp': 1768561294}
2026-01-16 12:01:35.975 INFO (paho-mqtt-client) [custom_components.zendure_ha.api] Topic: /***/SolarFlow 2400 AC/properties/report => {'timestamp': 1768561295, 'messageId': 226180, 'deviceId': '***', 'version': 2, 'product': 'solarFlow2400AC', 'properties': {'packNum': 1}, 'packData': [{'sn': '***'}]}
[...]

Issue #1136, opened 2026-02-26
[Bug] 200W Residual Power Problem


technoman63:
![Image](https://github.com/user-attachments/assets/d73d0f76-e054-4e74-a3fe-18887ab8f352)

### Home Assistant Version

2026.2.3

### Zendure Integration Version

1.2.5

### Describe the bug

Hello,

I'm having a problem. I have two Hyper 2000 solar panels and one SF800 Pro.

- The two Hyper panels are connected locally via Bluetooth.
- The SF800 Pro has the SDK.

On the SF800 Pro's off-grid outlet, I have my server rack, which consumes approximately 250W.

When I group three devices, I consistently end up with a grid return of around -200W during my home's solar production. I never get close to 0W, and neither the Hyper 2000 nor the SF800 Pro activates to collect the excess power.

If I connect the two Hyper 2000s individually, there's no problem; they reach 0W.

Is the problem with my off-grid outlet? Is there an issue with the three devices? Any suggestions on where I could test?

### To Reproduce

I don't know

### Expected Behaviour

to exhibit behavior close to 0W with my 3 devices

### What device are you using?

Hyper2000

### Log Extracts

```shell
not for the moment
```

jeje-it:
Hi, I have exactly the same bug on my side. It’s tracked here: https://github.com/Zendure/Zendure-HA/issues/1132
​

In my setup I also had a server  (~150–200W) on the SF2400ac off-grid outlet, and in smart_charging it caused a random residual grid injection/return (typically ~50–250W) instead of stabilizing close to 0W.
​

Validation test: if I turn off / remove the off-grid load, Smart Charge immediately works perfectly again and the Shelly grid power stabilizes around 0W.
​

Could you try the same test: run your 3-device group with no load on the off-grid outlet (or off-grid outlet disabled) and check if the grid return goes close to 0W?

technoman63:
Hello, indeed the problem is resolved, but we need to find a way to fix this, the off-network socket is not taken into account in the calculation.

jeje-it:
Thanks for the confirmation.
All that’s left now is to wait for a fix to be published in a future release.
On my side, I’ve moved my battery, and I’m no longer using the off-grid outlet.

Issue #1132, opened 2026-02-24
[Bug] SolarFlow 2400W - Incorrect power measurement in Smart Charge with offgrid consumption (Grid unstable)


jeje-it:
### Home Assistant Version

2026.02.2

### Zendure Integration Version

1.2.6-pre2

### Describe the bug

In "smart charging" mode (smart_charging), with stable 150-200W consumption on the offgrid outlet, the Shelly on the grid does NOT stabilize at ~0W despite stable solar production and stable house consumption.

Result: Completely random grid injection (50-250W) instead of precise regulation.

Validation test:
- ✅ Disable offgrid outlet → Smart Charge works perfectly, Shelly stabilizes at 0W
- ❌ Offgrid outlet 150-200W active → Power measurements faulty, regulation impossible

This is a power measurement bug specific to offgrid consumption in smart_charging mode.

### To Reproduce

1. SolarFlow 2400W in "smart charging" mode (smart_charging)
2. Connect stable 150-200W load to offgrid outlet
3. Stable solar production + stable house consumption
4. Observe Shelly grid → DOES NOT STABILIZE at 0W (random 50-250W injection)
5. Turn off offgrid outlet → Shelly immediately stabilizes at 0W

Tested setup:
- ShellyPro3EM on main grid
- Offgrid outlet with 150-200W load
- Stable solar, no consumption variation

### Expected Behaviour

Shelly grid regulation at 0W regardless of offgrid outlet state or load.

Expected behavior = observed behavior WITHOUT offgrid consumption (tested: perfect).

### What device are you using?

SF2400 AC

### Log Extracts

```shell
[...]

Issue #798, opened 2025-09-19
[Bug] Power Calculation wrong for SF2400AC


CommanderROR9:
### Home Assistant Version

2025.9.3

### Zendure Integration Version

1.14-pre7

### Describe the bug

Every version after 1.14-pre1 has been off by a certain margin for me when calculating the Power value. This leads to issues, both with Charging and Discharging in the Smart Matching and even Manual mode in Zendure Manager.

Today I finally figured out the source of the issue (I think).

What I did:
I took the value of the "power_actual" sensor and the "zendure_manager_power" and compared them. There was a discrepancy between those two values.
I created a Template and substracted one from the other.
The result was around 120W in my case, fluctuating slightly.
So I went searching for an explanation and found two possible Culprits. Offgrid Power and GridInputPower. Both were similar in Value, but it turns out that Grid Input Power was an exact match for the discrepancy.

This observation was made at a time when the battery was full and neither Charging nor Discharging. I will have to do some more tests in other scenarios to see if that changes anything, but wanted to present my initial findings for others to replicate.

### To Reproduce

Listed in description.

### Expected Behaviour

Exact matching.

### What device are you using?

SF2400 AC

### Log Extracts

```shell

```

CommanderROR9:
I have to admit that I don't really understand what is going on.
I did some brief testing with different settings.

If I enable Zendure Manager with Smart Matching, then it discharges into the Grid, although there is still loads of excess Solar. If I set it to manually discharge 400W, then the delta between power_actual and zendure_manager_power goes to 0, but somehow the Output isn't correct.
And if I turn Zendure Manager off and hand Control back to the Zendure HEMS, then everything immediately snaps back into place. The offset between power_actual and zendure_manager_power returns, but everything works correctly again with Smart Matching. I have no idea what kind of magic sauce Zendure is using...but it works and the Integration currently can't replicate it, at least on my System.

CommanderROR9:
The issue is still there obviously.

CommanderROR9:
Please keep it open. Thanks.

fellek:
take a look at https://github.com/Zendure/Zendure-HA/pull/1208

PR #1208, opened 2026-03-20
Fix: off-grid consumer power causes incorrect zero-feed-in setpoint #1015

fellek:
# Fix: off-grid consumer power causes incorrect zero-feed-in setpoint

Resolving #1015  
 
### Problem
 
When a consumer (e.g. fridge, router) is connected to the off-grid socket of an SF2400 AC while the battery is charging from ac connected solar power, the zero-feed-in calculation is incorrect. The grid feed-in permanently exceeds 0W by exactly the off-grid consumption.
 
**Root cause:** In `manager.py` line 425, `setpoint += home` uses the variable `home` which includes a `pwr_offgrid` offset (line 420). The off-grid socket supplies consumers *behind* the SF2400 AC — this power does not pass through the P1 meter at the grid connection point. The P1 meter sees the full `gridInputPower` (battery charging + off-grid supply combined), but the setpoint only corrects by `gridInputPower - gridOffPower`, under-accounting the device's actual grid draw.
 
**Example:** `gridInputPower=500`, `gridOffPower=+200` (fridge), `packInputPower=300`
- Current: `home = -500 + 200 = -300` → `setpoint += -300`, but P1 sees 500W → **200W unwanted feed-in**
- Fixed: `setpoint += -500` → matches P1 exactly → **0W feed-in** ✓
 
### Fix
 
Changed **only line 425**: use `-d.homeInput.asInt` (raw `gridInputPower`) instead of `home` for the setpoint adjustment.
 
Line 420 is unchanged — `home` still includes `pwr_offgrid` for correct charge/discharge/idle mode detection, and is still used in line 442 for the Manager Power entity.
 
### Verification
 
| Scenario | gridOff | setpoint (before) | setpoint (after) | Impact |
|---|---|---|---|---|
| Consumer 200W + charge | +200 | -300 ⚠️ | -500 ✅ | **Bug fixed** |
| Micro-inverter -300W + charge | -300 | -500 | -500 | Identical (`max(0,-300)=0`) |
| No off-grid + charge | 0 | -500 | -500 | Identical |
| Discharge / Idle scenarios | any | — | — | Line 425 not reached |

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -148,9 +148,9 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
 
         # initialize the api & p1 meter
         self.api.Init(self.config_entry.data, mqtt)
-        await self.update_fusegroups()
         self.update_p1meter(self.config_entry.data.get(CONF_P1METER, "sensor.power_actual"))
         await asyncio.sleep(1)  # allow other tasks to run
+        await self.update_fusegroups()
 
     async def update_fusegroups(self) -> None:
         _LOGGER.info("Update fusegroups")


fellek:
Solves issues:
#1132

#1136

fellek:
Solves #798

PR #1211, opened 2026-03-21
Change state_class for HA-entity total_kwh / energy to "total_increasing"

fellek:
Shows on every reload. The state class should be total or total_increasing instead of measurement. This is an entity configuration error.

`Entity sensor.zendure_manager_total_kwh is using state class 'measurement' 
which is impossible considering device class ('energy')`


FireSon:
I do not think that the current setup is wrong, because Capacity goes up/down as batteries are added/removed, therefore measurement is correct. please reopen if you still think it wrong!

fellek:
@FireSon you are totally right. I misinterpreted the entity name. `total_kwh` suggests being an entity for energymeter, which should be `total_increasing` . If I understand right, this entity is `total_capacity_in_kwh`. If possible I would suggest to rename this entity in the future.

edit: A  name could be: "installed_capacity_in_kwh"

PR #1221, opened 2026-03-22
Fix division by zero in power_charge and power_discharge

### LINKED ISSUE #1135: [Bug] Constant switches between Idle and Discharge
### Home Assistant Version

2026.2.1

1.2.5

### Describe the bug

When battery is at 100% charge in smart discharge mode it constantly switches from idle to bypass roughly every 30 seconds (see attached video).

https://app.filemail.com/d/nhvugdaxekesmyi

Here is the video couldn’t upload it here (only available for 7 days

### To Reproduce

When Soc has 100% and solar power is available and net consumption above 50w 

### Expected Behaviour

Stay on discharge and bypass don’t go to idle

SF800 Pro

### Log Extracts

```shell

```

### Other Information

_No response_

### LINKED ISSUE #1140: [Bug] Fusegroup output limit not respected
### Home Assistant Version

2026.2.3

v1.2.5

### Describe the bug

Thanks for this great project! It is already more capable than Zendure's cloud offering!

I have two Hyper 2000. "Hyper 2000 SE" set to "Fusegroup max 800w / 1200w charge", "Hyper 2000 SW" set to "Part of Hyper 2000 SE fusegroup".

I would expect the output power of the two Hypers to not exceed 800w. But that's not the case. When solar power is available and both Hyper batteries are full, there is no limit in output power whatsoever.

Maybe I am missing something and there is another way to achieve what I want? But the bug that the Fusegroup limit is not respected is quite obvious.

### To Reproduce

1. Assign  "Fusegroup max 800w / 1200w charge" to Hyper 2000 A
2. Assign "Part of Hyper 2000 A fusegroup" to Hyper 2000 B
3. Wait for the batteries of both Hypers to be fully charged on a sunny day
4. Watch the output power. The sum of the output of both Hypers is not limited to 800w - it outputs the full solar power of both Hypers.

### Expected Behaviour

The sum of the output of both Hypers is limited to 800w, regardless of how much input power is available.

Hyper2000

### Log Extracts

```shell

```

### Other Information

My goal is to limit the power export to the grid to 800w when the batteries are full and more than 800w are being produced. But I cannot get that to work. As an alternative, I tried to set "Energy Export" of the Hypers to "Forbidden" or "Disabled". That setting works, but it is too aggressive. I also have a SolarFlow 2400 AC (configured with its own Fusegroup) and would like its battery to be charged with excess solar power from the Hypers, but that does not happen.

### LINKED ISSUE #1199: [Bug] all devices start and stop when the batteries are full
### Home Assistant Version

2026.3.1

1.2.5

### Describe the bug

When the batteries are full and overcharging is prohibited in smart pairing mode, all devices (SF800+ and Hyper2000) regularly shut down and restart slowly after a few seconds, resulting in high network consumption

### To Reproduce

have all batteries full and solar panel directly in, and zendure manager in intelligent mod

### Expected Behaviour

i dont know

SF800

### Log Extracts

```shell
3 x sf800 and 4 hyper 2000
```

### Other Information

_No response_

### LINKED ISSUE #1203: [Bug] Smart power distribution mode not working once all batteries are fully charged and bypass forbidden
### Home Assistant Version

2026.3.2

V. 1.2.7 pre3

### Describe the bug

Describe the bug

2xHyper 2000, both with 3x AB2000X, Bypass forbidden, because I don't want to chase >2kW senselessly into the grid and burn up the inverters in the process.

Once, both Hyper reaches 100% SoC the output fluctuate between zero and X but never reaching needed output for the home grid.
(In addition manual power mode is no more working under these conditions, but does on V. 1.2.5)

I've filed this bug (or a very similar bug) in the past (#1151) which was closed, although I had reported the alleged solution as not working.

### To Reproduce

Use inverter and batterie at 100% SoC as described with solar surplus.

### Expected Behaviour

Distributing matching power to the home grid.

Hyper2000

### Log Extracts

```shell
Log will be inserted next day with 100% SoC and solar surplus.
Most likely tomorrow.
```

### Other Information

In real, the alleged solution worsened the situation by 
* output never reaching the power needed for the home
* manual mode not working anymore.

fellek:
I don't have multiple SolarFlows, so please verify this code, if related issues are fixed with that.

**Summary**

Charge_weight and discharge_weight in ZendureManager can become zero when devices reach 100% or 0% SOC, or when the accumulated weight drops to zero mid-iteration — causing a ZeroDivisionError that crashes the entire power distribution loop.

For charging (power_charge): when charge_weight == 0, all remaining devices are at 100% SOC — set pwr = 0 to skip charging.

For discharging (power_discharge): when discharge_weight == 0, distribute the remaining setpoint evenly across remaining devices so they can still pass through solar production                                                          
                                                                                                                                                                                                                                              
**Root cause** 

power_charge (line 513) and power_discharge (line 564) compute a weighted power share per device:                                                                                                                                           
   
`pwr = int(setpoint * (d.pwr_max * (100 - d.electricLevel.asInt)) / self.charge_weight)`                                                                                                                                                  
                  
When all devices in the loop have electricLevel == 100 (charge) or electricLevel == 0 (discharge), the weight sum is 0 and the division crashes. The exception is caught by the outer try/except in _p1_changed, which aborts the entire distribution cycle — leaving devices in their previous state and causing repeated start/stop cycling.
                                                                                                                                                                                                                                              **Related issues**  

possibly fixes #1203 — Smart power distribution not working at 100% SOC                                                                                                                                                                              
possibly fixes #1199 — All devices start and stop when batteries are full
possibly fixes #1135 — Constant switches between Idle and Discharge                                                                                                                                                                                  
possibly fixes #1140 — Fusegroup output limit not respected (partially)                                                                                                                                                                              
                                                                                                                                                                                                                                              
**Test plan** 

- [ ] Verify with multiple devices at 100% SOC: no crash, devices stay idle instead of cycling                                                                                                                                                  
- [ ] Verify with multiple devices at 0% SOC during discharge: power distributed evenly, no crash
- [ ] Verify normal operation (mixed SOC levels) still distributes power weighted by SOC as before                                                                                                                                              
- [ ] Verify SOCFULL devices still pass through solar production correctly                                                                                                                                                                      
- [ ] Verify fuse group limits are still respected during distribution

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -510,8 +510,18 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         limit = self.charge_limit
         setpoint = max(limit, setpoint)
         for i, d in enumerate(sorted(self.charge, key=lambda d: d.electricLevel.asInt, reverse=True)):
-            pwr = int(setpoint * (d.pwr_max * (100 - d.electricLevel.asInt)) / self.charge_weight)
-            self.charge_weight -= d.pwr_max * (100 - d.electricLevel.asInt)
+            # Weight per device: pwr_max * remaining capacity (100 - SOC%).
+            # Devices with lower SOC get a larger share of the charge power.
+            # Guard against division by zero: charge_weight can be 0 when all
+            # remaining devices are at 100% SOC (nothing left to charge) or when
+            # it drops to 0 mid-iteration after subtracting previous devices.
+            device_weight = d.pwr_max * (100 - d.electricLevel.asInt)
+            if self.charge_weight != 0:
+                pwr = int(setpoint * device_weight / self.charge_weight)
+            else:
+                # all remaining devices at 100% SOC — skip charging
+                pwr = 0
+            self.charge_weight -= device_weight
 
             # adjust the limit, make sure we have 'enough' power to charge
             limit -= d.pwr_max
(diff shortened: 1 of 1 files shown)


FireSon (approved):

sebisborn:
Sorry but it's not resolve #1199

PR #1228, opened 2026-03-23
security: opt-in MQTT user management and enforced local_only access

### LINKED ISSUE #1227: [SECURITY] Silent creation of insecure HA users (fixed passwords & remote access enabled)
### Home Assistant Version

2026.3.3

v1.2.7-pre3

## The Problem
The integration currently **silently** creates Home Assistant users for local MQTT communication without warning or consent. 

There are two major security concerns with this behavior:

1. **Fixed & Predictable Passwords**: The passwords for these users are fixed by the device firmware (MD5 of the `deviceId`) and cannot be changed. This makes them inherently weak if the `deviceId` is known.
2. **Remote Access Enabled**: These technical accounts are created with `local_only=False`. This means any attacker who discovers a `deviceId` could potentially use these credentials to log into the Home Assistant UI remotely (via Nabu Casa or a reverse proxy).

Users are not informed that these accounts are being created, nor are they given the choice to manage these credentials themselves.

### 1. Force `local_only=True` (Immediate Priority)
Technical accounts used solely for local MQTT traffic should **never** be allowed to authenticate from outside the local network. 
```python
user = await self.hass.auth.async_create_user(deviceId, group_ids=[GROUP_ID_USER], local_only=True)
```

### 2. Change to Opt-in Management
The automatic user creation should be an **opt-in** feature in the integration settings. By default, no HA users should be created unless the user explicitly enables this option during or after setup.

### 3. Documentation for Manual Setup
Clearly explain how to manually set up a secure local user for those who do not use the opt-in feature:
- **Username**: Must be the `deviceId`.
- **Password**: The MD5 hash of the `deviceId`.
- **Constraint**: The user should be restricted to **"Local access only"** in HA settings to prevent remote authentication.

### To Reproduce

   1. Install the Zendure-HA integration.
   2. During configuration, provide a local MQTT server address (e.g., the IP of your Home Assistant instance).
   3. Complete the setup and let the integration load your devices.
   4. Go to Settings > People > Users.
   5. Observe that new user accounts named after your Zendure deviceId have been created automatically.
   6. Note that no notification or confirmation was presented during the process, and these users are not restricted to "Local access only" by default.

### Expected Behaviour

  Expected Behaviour
   1. Security by Default: Any technical account created for MQTT must be restricted to local_only=True to prevent unauthorized remote UI authentication.
   2. Contextual Logic: The integration should only attempt to create/manage HA users if the official Mosquitto add-on is detected as installed and running.
   3. Opt-in Feature: User management should ideally be an Opt-in setting (disabled by default) to respect the user's control over their authentication system.
   4. Documentation: The README or official documentation must explicitly mention this behavior, explaining why these accounts are created, that their passwords are fixed (firmware MD5), and how to set them up manually if the automated feature is not used.

SF2400 AC

### Log Extracts

```shell

```

### Other Information

_No response_

thib3113:
# Description
This PR addresses several security concerns regarding how the integration manages technical Home Assistant users for local MQTT communication.

### Changes:
1. **Enforced `local_only=True`**: All technical MQTT users created by the integration are now strictly restricted to local network access. This prevents unauthorized remote UI authentication using these technical credentials.
2. **Opt-in Management**: Automatic user creation is now an opt-in feature. Users must explicitly enable the **"Automatically manage MQTT users"** option during setup or in the integration options.
3. **Complete Translations**: Added missing translations for the MQTT configuration step and the new security option

### Why this is important:
Zendure devices use fixed MQTT passwords (MD5 of the `deviceId`). By forcing `local_only` access and requiring explicit consent, we significantly reduce the attack surface and give users better control over their Home Assistant authentication system.

---

### MQTT Security & User Management

When using **Local MQTT** with your Zendure devices, technical user accounts are required for authentication.

#### 1. Automatic Management (Recommended for HAOS)
You can let the integration manage these users for you by enabling **"Automatically manage MQTT users"** in the settings.
- **Security**: The integration will automatically set these users to **"Local access only"**.
- **Requirement**: This works best with the official **Mosquitto broker addon**.

#### 2. Manual Management (External Brokers)
If you manage your own MQTT broker or prefer manual control, keep the automatic option disabled and create users manually:
- **Username**: Your device's `deviceId` (e.g., `HEC4...`).
- **Password**: The **UPPERCASE MD5 hash** of the `deviceId`.
- **Recommendation**: If using Mosquitto addon, manually enable the **"Can only log in from the local network"** toggle for these accounts in HA Settings > People > Users.

Fix #1227

### DIFF
--- a/custom_components/zendure_ha/config_flow.py
+++ b/custom_components/zendure_ha/config_flow.py
@@ -117,10 +117,8 @@ class ZendureConfigFlow(ConfigFlow, domain=DOMAIN):
         schema = self.data_schema
         if user_input is not None:
             self._user_input = self._user_input | user_input
-            use_mqtt = self._user_input.get(CONF_MQTTLOCAL, False)
-            
-            if use_mqtt and user_input.get(CONF_MQTTSERVER) is None:
-                # User checked "Use local MQTT", show the MQTT step
+            use_mqtt = user_input.get(CONF_MQTTLOCAL, False)
+            if use_mqtt:
                 schema = self.mqtt_schema
             else:
                 try:
@@ -134,11 +132,6 @@ class ZendureConfigFlow(ConfigFlow, domain=DOMAIN):
                     self._abort_if_unique_id_mismatch()
 
                     return self.async_update_reload_and_abort(entry, data=self._user_input)
-        else:
-            # Pre-fill with existing data
-            self._user_input = dict(entry.data)
-            if self._user_input.get(CONF_MQTTLOCAL):
-                schema = self.mqtt_schema
 
         return self.async_show_form(
             step_id="reconfigure",


FireSon (approved):
Small items fixed

PR #1247, opened 2026-04-04
async adder fixes error on startup

zaubara:
fixes commit 2c46d25


FireSon:
@zaubara thanks! I did not think the change through properly. I changed it back to the original order, which should be sufficient for now. Can you check and reopen the PR is it is still not working?

PR #1272, opened 2026-04-21
Fix renaming

harrymayr:
With these changes, the renaming of the entities and batteries works well. Also all helper sensors and automations work.
Still Node-Red may need a check with respect to the entities.


FireSon:
Thanks for the work on this, @harrymayr !!

I've opened #1275  which includes all five of your commits unchanged plus a small cleanup on top addressing a few items from review (`@staticmethod` on `get_battery_type`, `change_id` recursion propagation, a few minor hygiene fixes, and restoring `hw_version` for fresh installs). Your authorship is preserved on the original commits.

PR #1273, opened 2026-04-22
zero-export Power Distribution example

LukasQ:
If you're controlling the power management through HA, this PR documents a working automation for anyone else trying to keep their SolarFlow 800 Pro from cheerfully feeding power back to the grid.
  - Tunable `feed_buffer` (default 40 W) and `dead_band` (default 10 W), with actual explanations of what those words mean                                                                                                                                                  
  - Availability guards on grid sensor, battery SoC, and `output_limit` — so it doesn't compute a new setpoint from a phantom zero if the device is sulking
  - LiFePO4-friendly SoC taper between `soc_min` and `soc_min + 10%`
  - Placeholders (`<GRID_POWER_SENSOR>`, `<ZENDURE_DEVICE>`) so the snippet is copy-pasteable without everyone learning what I named my device

Pure docs. No code touched. No entities harmed in the making of this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ This Home Assistant integration connects your Zendure devices to Home Assistant,
 
 - **Device Automation:**
   - Cheap hours.
+  - [Example: Zero-Export Power Distribution](#example-zero-export-power-distribution)
 
 ## Minimum Requirements
 - [Home Assistant](https://github.com/home-assistant/core) 2025.5+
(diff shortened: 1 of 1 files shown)


FireSon:
@LukasQ Thanks for the PR! I do not really know what the added value of your PR should bring, on top of the Zendure Manager, which already can do zero-export?

LukasQ:
well... I thought of more fine granular config. But I totally forgot about the manager.

PR #1280, opened 2026-04-23
Api refactoring

fellek:
## Summary
Extracts API layer responsibilities (authentication, MQTT protocol, device transport) 
into separate, testable modules. Applies 6 critical bugfixes for thread leaks, 
error handling, and shared state issues.

## Changes
- New modules: `api_auth.py`, `mqtt_protocol.py`, `zendure_sdk.py`
- Bugfixes: MQTT thread leaks, API error handling, manager state, coordinator loop, file paths

## Related
Depends on: Nothing
Followed by: BLE extraction

### DIFF
--- a/custom_components/zendure_ha/__init__.py
+++ b/custom_components/zendure_ha/__init__.py
@@ -41,7 +41,7 @@ async def update_listener(_hass: HomeAssistant, entry: ZendureConfigEntry) -> No
     """Handle options update."""
     _LOGGER.debug("Updating Zendure config entry: %s", entry.entry_id)
     Api.mqttLogging = entry.data.get(CONF_MQTTLOG, False)
-    ZendureManager.simulation = entry.data.get(CONF_SIM, False)
+    entry.runtime_data.simulation = entry.data.get(CONF_SIM, False)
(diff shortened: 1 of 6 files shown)


FireSon:
@fellek Not really sure what you want to archive. At this point in time I do not want major reconstruction of the integration.

fellek:
Hi @FireSon, thanks for the feedback — totally fair.

I had opened a discussion in [#1244](https://github.com/Zendure/Zendure-HA/discussions/1244) beforehand to align on scope — feel free to continue there if useful. 

The structural side (new files, module split) is optional. The part I'd really like to get in are the 6 bugfixes that were found during the cleanup:

- MQTT clients not stopped on shutdown → thread leaks
- API returning `None` silently on non-200 responses
- `devices`/`fuse_groups` as class variables → shared across coordinator instances
- `_schedule_refresh()` called from within `_async_update_data()` → refresh loop
- `Init()` called on every reconnect → duplicate MQTT subscriptions
- Hardcoded `simulation.csv` path instead of `hass.config.path()`

I'm happy to open a separate PR with just those fixes, no structural changes at all. Would that work?

And whenever you feel the time is right for a larger refactor, I'd love to be looped in — happy to revisit then.

Issue #1263, opened 2026-04-16
[Bug] Zendure Token: invalid input - no Devices found


deady2000:
### Home Assistant Version

2026.4.2

### Zendure Integration Version

v1.2.5

### Describe the bug

Dear ZHA Team,

I am struggling with the first step of the Zendure integration in Home Assistant using my Solarflow 800 Pro 2. Every time I enter my Cloud Token, I get the following error in the log:

Zendure API does not Reply Any Devices: {'code': 200, 'success': True, 'data': {'mqtt': {}, 'deviceList': []}, 'msg': 'Operation success'}

I'm a new user, never had any other Solarflow hardware or an account before, so everything is fresh setup.
I only have one account, that is my main and only account. My location is Europe/Germany.

The connection seems to work (no auth error), but the device list remains empty. Here is what I have tried so far without success:

- Restarted Home Assistant and the  Solarflow 800 Pro 2 multiple times.
- Checked the HA system time, 100% in sync with radio clock, smartphone, PC and installed chrony for testing
- Reinstalled the Zendure integration several times, including various pre-release versions.
- Verified the token. Entering a wrong token on purpose results in a connection error, so the current one is valid. The token length is 52 characters.
- Deleted the  Solarflow 800 Pro 2 from HEMS, tried to configure ZHA (same error) and re-added it in the Zendure app tried to configure ZHA again (same error).
- power cycled the  Solarflow 800 Pro 2
- Changed the Zendure app language to English and set the time zone to USA for testing purposes.
- Tried the setup from different machines (Windows, IOS, and Mac).
- I'm not able to enable debugging as I can't configure the integration

Is there maybe a region problem or is the Solarflow 800 Pro 2 unsupported?

Many thanks, Patrick

### To Reproduce

Add integration
Fill all fields
Click on "OK"
Error "invalid input"
Check Logfile - Zendure API does not Reply Any Devices: {'code': 200, 'success': True, 'data': {'mqtt': {}, 'deviceList': []}, 'msg': 'Operation success'}

### Expected Behaviour

Successful Solarflow 800 Pro 2 integration in my HA setup, Im not planning to control it with ZHA, only monitoring for the energy data.

### What device are you using?

SF800 Pro

### Log Extracts

```shell
Zendure API does not Reply Any Devices: {'code': 200, 'success': True, 'data': {'mqtt': {}, 'deviceList': []}, 'msg': 'Operation success'}
```

harrymayr:
[...]

Issue #1256, opened 2026-04-11
[Bug] Testing the ACE 1500 + AB2000X: Debug Logs and Discharge Logic


Popoboxxo:
### Home Assistant Version

2026.3.4

### Zendure Integration Version

Version 1.2.5

### Describe the bug

Hey everyone,

Referring  #844  (can't reopen the issue)

I got another AB2000X after an Aliexpress sale – and finally I'm able to give you some debug logs:

2026-04-11 20:31:39.904 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-3 isFast:False, setpoint:751W stored:0W
2026-04-11 20:31:39.904 INFO (MainThread) [custom_components.zendure_ha.manager] Discharge => setpoint 751W
2026-04-11 20:31:39.904 INFO (MainThread) [custom_components.zendure_ha.device] Power discharge Hyper 2000 OG => no action [power 375]
2026-04-11 20:31:39.904 INFO (MainThread) [custom_components.zendure_ha.device] Power discharge Hyper 2000 EG => no action [power 379]
2026-04-11 20:31:39.904 INFO (MainThread) [custom_components.zendure_ha.devices.ace1500] Power discharge Ace 1500 => 50
2026-04-11 20:31:39.904 ERROR (MainThread) [custom_components.zendure_ha.manager] name 'sp' is not defined
**2026-04-11 20:31:39.906 ERROR (MainThread) [custom_components.zendure_ha.manager] Traceback (most recent call last):
  File "/config/custom_components/zendure_ha/manager.py", line 398, in _p1_changed
    await self.powerChanged(p1, isFast, time)
  File "/config/custom_components/zendure_ha/manager.py", line 457, in powerChanged
    await self.power_discharge(setpoint)
  File "/config/custom_components/zendure_ha/manager.py", line 578, in power_discharge
    await d.power_discharge(SmartMode.POWER_START)
  File "/config/custom_components/zendure_ha/device.py", line 498, in power_discharge
    return await self.discharge(power)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/zendure_ha/devices/ace1500.py", line 57, in discharge
    "outPower": max(0, power + sp),
                               ^^
NameError: name 'sp' is not defined**

Since the ACE 1500 isn't capable of feeding into the grid anyway, it would be enough if it just enables discharging. 

[...]

PR #1282, opened 2026-04-23
Several fixes and backdoor

harrymayr:
This PR cover several things:

- empty device list for SF 800 pro 2 and Superbase V4600 with token only. Could be solved by using the App-email and App-PW together with the token, seperated by colon -> token:email:PW -> #1263 and older issues
- Changing byPass to a sensor, because it covers 4 states (`Off`, `On SoC limit`, `On min Soc` and `grid direct supply`)
- Removing  AC charge settings for HUB family. Trying to start AC charge for a HUB (with -50W) leed to strange behavior (#1189)
- Adding an additional pack_state for the ACE 1500 (USV) -> #1256
- Avoid stopping of discharge for bypassed devices on charge -> #1243
- Keep smartmode at 1 if offGridPower > 0
- Add random-value (0-10) to start-power, to avoid non reaction of SF 800 pro in rare cases -> #1283


fellek:
**Observation: mqtt-value `pass` only carries values 0, 2, and 3 — state 1 is unobserved**

Based on MQTT captures across multiple devices (SolarFlow 800, SF 2400), the `pass` field only ever carries `0`, `2`, or `3`. Value `1` has never been observed in practice.

I've documented the three active states in detail [here](https://github.com/fellek/zendure-ha/blob/refactor/device-structure-optimizing/docs/ZenSDK_Docs_merged_2026-04-05.md#pass--bypass-modus-detailwerte):

| Value | Name | Direction |
|-------|------|-----------|
| 0 | BYPASS-OFF | Normal operation |
| 2 | BYPASS-REVERSE | Surplus from offgrid socket / PV → grid |
| 3 | BYPASS-INPUT | Grid → offgrid socket (battery bypassed) |

This creates a problem in the `manager.py` change:
```python
if d.byPass.asInt == 1:
    continue
```
Since `pass=1` is never sent by the firmware, this guard never fires. Devices in BYPASS-REVERSE (2) or BYPASS-INPUT (3) will still receive `power_discharge(0)`, which defeats the fix for #1243. The condition should be `d.byPass.asInt in (2, 3)`.

**On the "UPS" label for `packState=3` (ACE 1500):**
At least on the **SF2400AC**, stable power at the offgrid socket cannot be guaranteed — power outages at the GridOff socket have been reported specifically during bypass transitions (`pass=3`, BYPASS-INPUT), as documented in #1016. The relay switchover causes a measurable interruption; sensitive loads do not survive it. Calling this state "UPS" (Uninterruptible Power Supply) is therefore misleading, regardless of device model. A label like **"Backup power"** or **"Grid direct supply"** would be more accurate. Whether the ACE 1500 behaves differently would need separate verification before "UPS" can be claimed.

harrymayr:
I have logs at least from a HUB 1200 with pass==1. Need to check my other logs to check for pass==1 on other devices.
The packstate==3 was reported from the user in #1256 and also that the App show the status UPS
And I never see packstate==3 on my Hyper, Sf 800pro and SF 2400 AC

fellek:
This might be true for ACE series and would be new finding. 

There is no `packState=3` in official documentation, but `pass=2|3` neither.

| Attribute | Type | Access | Unit | Range | Description |
|-----------|------|--------|------|-------|------------|
| packState | int | RO | — | 0-2 | Pack state |
| pass | int | RO | — | 0-1 | Pass-through state |

In my case `packState` (SF2400AC) transports these information:

| Attribute | Type | Access |  Description |
|-----------|------|--------|------------|
| packState | int | RO | 0: Standby, 1: Charge, 2: Discharge | 

If `packState=3` means real UPS, that would be great for ACE users!
If the ACE Series has `pass=1`, it needs to be clarified whether this is a collective state for bypass, or whether other values are also provided.
In the first case, `pass=1` means `bypass=On` (in both directions) and would not be a breaking change for devices that indicate the bypass direction with `pass=2|3`.

custom_components/zendure_ha/translations/en.json
```diff
@@ -289,7 +289,7 @@
           "0": "Off",
           "1": "On SoC limit",
```
fellek: Just make it `1=on`

fellek: In my findings:
The `pass` value describes the bypass relay state, not the reason for it. The firmware activates bypass independently of SOC — what matters is the power flow at the offgrid socket or PV-MPPT ports:
 - `pass=1`: Legacy boolean value (bypass active, direction unknown) — likely from older firmware versions
 - `pass=2`: Bypass in feed-in direction — energy flows from the offgrid socket into the house (e.g. a micro-inverter or PV input via MPPT ports), and (manual) charging is lower than available energy. SOCFULLmets this rule by charging=0
 - `pass=3`: Bypass in consumption direction — grid power flows through to the offgrid socket (e.g. a connected load is being supplied from the grid). This occurs when the (manually) configured discharge power is lower than the offgrid load demand — the firmware takes over and sources the deficit directly from the grid. SOCEMPTY mets this rule by discharging=0
 
SOC is therefore not a direct trigger for bypass. The firmware switches the relay based on the detected power flow at the offgrid port/PV-MPPT. A SOCEMPTY device with active bypass (pass=3) routes grid power to the offgrid load — the battery itself remains passive. A SOCFULL device with pass=2 routes energy fed in from the offgrid port into the house, without loading the battery.

harrymayr: > * This occurs when the (manually) configured discharge power is lower than the offgrid load demand — the firmware takes over and sources the deficit directly from the grid. SOCEMPTY mets this rule by discharging=0

This is not true wrt of my data. pass==3 happens only, if **all** offGrid power comes from the gridInput. This you can also see in the App. It only shows `direct grid supply` (pass=3) if charge/discharge power is set to 0, which means, that all power comes from grid.
This is also the reason, that it doesn't matter on pass==3, if you stop discharge, because power is already set to 0.
For pass==2 I think you are right, but I will check first the different logs I have collected over the time. With pass==1 you are correct, I find it only for the HUB family.

fellek: this sounds interesting, I will check `pass==3` again with your parameters!

harrymayr: > SOC is therefore not a direct trigger for bypass.

Sorry, but SOC is the only trigger for pass-states 0 to 2, because it controls (together with minSoc and socLimit) if there is a charge/discharge of the batteries. And if there is no charge/discharge of the batteries, but power from or to the grid, bypass will be set.
Only pass state 3 is controlled by the power which need to be set to 0 with offGridPower > 0.

FireSon:
Hallo Harry  thanks for putting this together. 

The problem for me is that it is too many changes in one PR.  Could we split it?

Following items are clearly correct:
* solarflow 800 pro2 → SolarFlow800Pro mapping in api.py
* Trailing newlines on en.json / fr.json / nl.json
* homeInput default of 0 in device.py

These items I think are not:
* HUB 1200 / 2000 AC /ACE1500 Why not set the limit for the Hub to 0 if it cannot charge?
* Bypass binary → multi-state migration. I do not understand why we need this? We already use SOCFULL bypass protection.
* ZenSdk smartMode change (device.py:741-754). Not sure but if we use self.offGrid.asInt it will raise exceptions if the device does not have offGrid. But more importantly, using smartmode == 0 will write into the permanent memory of the device, which CANNOT handle very frequent updates, and will lead to the device being unresponsive! I understand that smartmode == 1 has problems for offgrid, but it might be better to exclude devices with offgrid != 0 from the distribution. 
* #1283 fix — random jitter on start_pwr. I traced the manager loop, devices that fail to start are already retried each tick because powerChanged() could this behaviour be caused by writing smartmode = 0? As far as I can see, this will not solve it.
* Zendure made the authentication we are using now, especially for the integration. I do not think we can go this route!! However, there is a file where the integration stores it devicelist (config/.storage/zendure_ha.storage. This file can be manually edited, all devices from this file are added to the device list. Can you try adding the necessary data to it?

Thanks,
Jan

custom_components/zendure_ha/translations/en.json
fellek: Thanks for clarification. Of course `bypass==off` in every case `battery_power <> 0`.  
I have done some tests with pv-microinverter plugged in offgrid socket. Results are visible in #1285. besides charging and discharging does not work in this setup properly, in some of these cases `bypass==on` without SOCFULL or SOCEMPTY, just by changing Manager-Mode. I think, its intentional, especially in cases, where `battery_power==0` is forced by Mode-OFF. 

For `bypass==3` I have mixed-up different things. your are right, bypass is only triggered on battery_power==0 and offGridPower > 0.  

In case there is an integrated MPPT + OffGrid-Socket + AC-Connector to the Grid (like new SF2400AC Pro and others) , there might be situation of MPPT Power >0 with Offgrid  >0 usage, but charging forbidden. in this case battery_power==0, offGrid > 0, but bypass==3 to feedin surplus energy, unless export is not forbidden.

can you confirm this theoretical case? I am unable to test it.

fellek:
> ZenSdk smartMode change (device.py:741-754). Not sure but if we use self.offGrid.asInt it will raise exceptions if the device does not have offGrid. But more importantly, using smartmode == 0 will write into the permanent memory of the device, which CANNOT handle very frequent updates, and will lead to the device being unresponsive! I understand that smartmode == 1 has problems for offgrid, but it might be better to exclude devices with offgrid != 0 from the distribution.

Hi @FireSon, 
do you have a roadmap, what fixes (supported devices, fixed bugs) should be done in 1.2.7 release? I would like to support, making it stable.

harrymayr:
Added 2 new PR, without the old auth.

FireSon:
> Hi @FireSon, do you have a roadmap, what fixes (supported devices, fixed bugs) should be done in 1.2.7 release? I would like to support, making it stable.

Hi @fellek at this moment I am working on a new version which basically refactors everything. That is the main reason I am pushing back on major refactoring of the current integration. The migration 'problem' was the main reason for not releasing 1.2.7, and since it took so long to fix it, many new changes were added. At this moment is seems that the migration issues are fixed and once that is confirm we can release. 
Are there any specific devices and bugs you are referring to?

fellek:
@FireSon Thanks for the update on the current project status. I understand that the ongoing bug fixes are blocking a larger refactoring, and I would like to help ensure we reach a stable version 1.2.7 if there is anything I can contribute. 

In my opinion, the integration's codebase is hitting its limits in many places due to its complexity (and partly due to missing information regarding the firmware/API). That is why I started working on a refactoring a few weeks ago; I reworked the integration using a stricter OOP approach and added unit tests and documentation. 

Aside from the bugs mentioned in #1280, I currently don't have any bugs that I expect can be resolved in 1.2.7. Please let me know if I can help with the release and whether collaborating on the refactoring is an option.

FireSon:
@fellek Well I do not agree that the current codebase is bad. It is easy to overengineer OOP, especially using AI tools. I did not see the discussion you mentioned, but I am not looking for AI to tell me how to structure my code and to fragment it into many file where humans loose the overview. 
Not sure what you/AI calls bugs, often times it is not really a bug if you push back. So can y point out what is actually not working properly?

fellek:
Hi @Username, 

I hope my previous message didn't give the impression that I'm unhappy with the code. My observation is simply that the wide variety of Zendure devices, with their different inputs and outputs, often requires specific handling to achieve the desired behavior. 

In the case of the SF2400AC, I am currently working on some improvements that fit my specific setup: 

- Handling wake-up after 'deep' standby. 
- Addressing the Bypass/Offgrid/SmartMode issue regarding power stability (there is potential competition with device types w/o offgrid: https://github.com/Zendure/Zendure-HA/pull/1282#issuecomment-4319205722). 
- Preventing battery drain despite SOCEMPTY. 

I completely agree that OOP isn't a universal solution, and it can definitely lead to over-engineering. However, provided we keep that risk in mind, I find the approach of data encapsulation (read/write) and the separation of program logic very helpful. I am aware that everyone has their own personal preference regarding OOP, and there are good arguments for both approaches.

Regarding the fixes I mentioned in https://github.com/Zendure/Zendure-HA/pull/1280#issuecomment-4308908228 , I can offer to provide more details in the next few days. In most cases, these will be "invisible" improvements that mainly focus on explicitly defining states.

I certainly have no reason to criticize the codebase. For me, it is a fantastic foundation, especially in such an active repository and the main reason I decided to switch to Zendure. Personally, I work well with OOP and prefer writing encapsulated, focused functions/methods. This helps contain specific logic and avoids unintended side effects with other parts of the program.

Let me know what you think about this approach or if I missed anything.

FireSon:
@fellek Well no problems, and any help or insights are always valuable! It is just that I have a certain idea behind the architecture, which might not always be plain to see. Also fair to say, that other scenarios exists, which I am not aware of. I think OOP is very well implemented in the current codebase, and in order to keep it as simple as possible (I am a big fan of KISS!) you have to make choices. 
The thing I am working on is basically a complete new approach of solving all issues related to energy management. If you have time, and like to develop new things, you could join me on that. It will also need solutions to the problems you mentioned. 
In any case it would be great to get more info on the fixes you mentioned!

PR #1287, opened 2026-04-25
Hub and ACE fixes

harrymayr:
- Removing AC charge settings for HUB family. Trying to start AC charge for a HUB (with -50W) leed to strange behavior (https://github.com/Zendure/Zendure-HA/discussions/1189)
- Adding an additional pack_state for the ACE 1500 (USV) -> https://github.com/Zendure/Zendure-HA/issues/1256

### DIFF
--- a/custom_components/zendure_ha/devices/hub1200.py
+++ b/custom_components/zendure_ha/devices/hub1200.py
@@ -14,25 +14,20 @@ class Hub1200(ZendureLegacy):
     def __init__(self, hass: HomeAssistant, deviceId: str, prodName: str, definition: Any) -> None:
         """Initialise Hub1200."""
         super().__init__(hass, deviceId, prodName, definition["productModel"], definition)
-        """Can control the ACE 1500 with max 900W AC inputPower"""
-        self.setLimits(-900, 1200)
+        self.setLimits(0, 1200)
         self.maxSolar = -800
 
-    def batteryUpdate(self, batteries: list[ZendureBattery]) -> None:
-        # Check if any battery has kWh > 1
-        if any(battery.kWh > 1 for battery in batteries):
-            self.powerMin = -1200
-            self.limitInput.update_range(0, abs(self.powerMin))
-
     async def charge(self, power: int) -> int:
-        _LOGGER.info("Power charge %s => %s", self.name, power)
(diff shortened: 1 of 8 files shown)


custom_components/zendure_ha/devices/hub1200.py
```diff
@@ -14,25 +14,20 @@ class Hub1200(ZendureLegacy):
...
-        _LOGGER.info("Power charge %s => %s", self.name, power)
+        _LOGGER.info("AC Power charge %s not available => set power from %s to 0", self.name, power)
+        # The HUB family does not have AC charging possibility (even with ACE 1500), so set it to idle
+
         self.mqttInvoke(
             {
-                "arguments": [{"autoModelProgram": 2, "autoModelValue": power, "msgType": 1, "autoModel": 8}],
+                "arguments": [{"autoModelProgram": 0, "autoModelValue": 0, "msgType": 1, "autoModel": 0}],
                 "function": "deviceAutomation",
             }
```
FireSon, line 28: Why do we need to send this command if it does not work?

custom_components/zendure_ha/translations/de.json
```diff
@@ -273,7 +274,8 @@
         "state": {
           "0": "Aus",
           "1": "An",
-          "2": "Automatisch An"
+          "2": "An",
```
FireSon, line 277: Why not Automatisch An anymore?

custom_components/zendure_ha/devices/hub1200.py
harrymayr: HUB has no AC charging capabilities, but of course can discharge. But if charge is requested, no discharge should happen, so set it to idle.

custom_components/zendure_ha/translations/de.json
harrymayr: It could be, that if you have the possibility to switch ON the Bypass, as it is (as I know) for a HUB, pass is 1 and 2 indicate, that PV goes direct into home grid. But I'm not sure, if it is really "automatic". We can keep it unchanged, but to be honest, I do not really know, what I should understand what "automatic on" does mean. 
Finally it makes no difference.

PR #1288, opened 2026-04-25
Fixes for off grid devices

harrymayr:
- Add SF 800 pro 2 (may not show up with token auth.)
- Changing byPass to a sensor, because it covers 3 states (`Off`, `On `and `grid direct supply`)
- Avoid stopping of discharge for bypassed devices on charge -> https://github.com/Zendure/Zendure-HA/issues/1243
- Keep smartmode at 1 if offGridPower > 0
- Add random-value (0-10) to start-power, to avoid non reaction of SF 800 pro in rare cases -> https://github.com/Zendure/Zendure-HA/issues/1283


fellek:
confirm fixed bug https://github.com/zendure/zendure-HA/issues/1016 no power outage in multiple charge-discharge-bypass transitions.

Issue #1235, opened 2026-03-29
[Bug] SF800Pro, SoC Full triggers oscillating of output


CounterChip:
### Home Assistant Version

2026.3.4

### Zendure Integration Version

1.2.7-pre3
edit:
also 1.2.7-pre4

### Describe the bug

When SoC reaches state full, the output power is oscillating.

### To Reproduce

Fresh install, first time user of the Zendure-HA integration
Power Meter is a Shelly Pro3EM.
A combination sensor adds the separate power phases.

Then a user definable Target Value is added, the result is the power input for the P1 sensor.
(Basicly to set a fixed grid offset value and for testing of the regulation functionality)

1x SF800Pro installed

### Expected Behaviour

I expect that the regulation works like it does when the SoC is not full.
Basicly if the SoC isnt full, it works pretty decent and i would like to see a similiar result, when SoC reaches its full state.

### What device are you using?

SF800 Pro

### Log Extracts

```shell

```

### Other Information

I already did some research myself and did some calculations, so i think there happend some code changes, which together resulted in this oscillation bug.
I already tried the [PR#1220](https://github.com/Zendure/Zendure-HA/pull/1220) and  [PR#1221](https://github.com/Zendure/Zendure-HA/pull/1221) which both didnt solve the problem.

Example of seeing it first time:
Starts when SoC full is triggered, with the oscillation.

<img width="1000" height="448" alt="Image" src="https://github.com/user-attachments/assets/f1916b95-e62e-482d-8204-5e58c9099900" />

 I tried to get another example and then stopped it, thats why the diagram ends during the plot:

<img width="1345" height="411" alt="Image" src="https://github.com/user-attachments/assets/f9f14afe-5111-4346-a890-4cd858027d6c" />

**So thats the problem, below follows some try and errors by myself to find a solution:**

I read the code and tried to understand it. I made some calculations and found out, that if i take the P1 Sensor value 2-3 seconds after the setpoint change of the inverter i would get somewhere near 0 for the correct amount of power.

Spreadsheet:
Ausgangsleistung means inverter output power
Solarleistung means solar input power
<img width="1092" height="619" alt="Image" src="https://github.com/user-attachments/assets/55f58077-2d92-4d46-af60-ebe4919fd729" />

So there must be something, which prevents the correct power to stay where it should be.

I tried to fix this for myself and it worked after both changes, only one of these two wasnt sufficient:

[...]

PR #1296, opened 2026-04-28
should fix p1 matching when SOCFULL, but grid reverse is disabled

zaubara:
ie. request only whats required from the solar panels, when theres more solar available then required

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -125,6 +125,7 @@ class ZendureDevice(EntityDevice):
         self.pwr_produced: int = 0
         self.actualKwh: float = 0.0
         self.state: DeviceState = DeviceState.OFFLINE
+        self.exports_bypass: bool = True
 
         self.create_entities()
 
@@ -244,6 +245,8 @@ class ZendureDevice(EntityDevice):
                         if self.electricLevel.asInt == 100:
                             self.nextCalibration.update_value(dt_util.now() + timedelta(days=30))
                         self.availableKwh.update_value((self.electricLevel.asNumber - self.minSoc.asNumber) / 100 * self.kWh)
+                    case "gridReverse":
+                        self.exports_bypass = value != 2
         except Exception as e:
             _LOGGER.error("EntityUpdate error %s %s %s!", self.name, key, e)
             _LOGGER.error(traceback.format_exc())
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -428,6 +428,12 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                 # get power production
                 d.pwr_produced = min(0, d.batteryOutput.asInt + d.homeInput.asInt - d.batteryInput.asInt - d.homeOutput.asInt)
                 self.produced -= d.pwr_produced
+                grid_reverse_entity = d.entities.get("gridReverse")
+                grid_reverse = grid_reverse_entity.value if isinstance(grid_reverse_entity, ZendureSelect) else None
+                # SOCFULL bypass power should only be accounted when export is allowed
+                # (or the device does not expose gridReverse). With gridReverse=forbidden,
+                # produced solar follows requested output and must not be subtracted again.
+                socfull_bypass = d.state == DeviceState.SOCFULL and grid_reverse in (None, 1, "allow")
 
                 # only positive pwr_offgrid must be taken into account, negative values count a solarInput
                 if (home := -d.homeInput.asInt + max(0, d.pwr_offgrid)) < 0:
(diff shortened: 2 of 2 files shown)


CounterChip (commented):
Works with SoC==Full and a SolarFlow 800 Pro, with this PR.
Tried small appliences in the screenshot below for small jumps and a boiler to see if it regulates correctly after a big change. (Boiler from 16:19 to 16:21)

Probably Fixes Bug #1235 and probably some others, but havent yet testet other states, only at SoC==Full.

Image

edit:
Discharge works also fine.

FireSon:
@zaubara Thanks for the PR! At this moment there are 3 different PR's handling gridReverse=forbidden. They all solve different parts if I understand it correctly. 
#1296 (zaubara) — fixes how discharge_bypass is accumulated (don't accumulate for forbidden devices)
#1318 (ahocevar) — fixes how discharge_bypass is consumed in the setpoint calculation
#1330 (ahocevar) — fixes how forbidden devices affect dev_start and pwr_low hysteresis in power_discharge

In general, I do want to keep the d.state == DeviceState.SOCFULL as the check. But if that is not sufficient, and you use socfull_bypass why not fix it in the state? That does not break my intention of the state.

zaubara:
I‘m not quite sure what you propose to fix in the state? SOCFULL sounds like it should apply when the battery is full - but I don’t think it should handle anything differently (socfull or not) when the export is disabled.

FireSon:
@zaubara pushed a small restructure (`ecb87b5`) before merging.

What it does:
- Adds `exports_bypass: bool` as an instance attribute on `ZendureDevice` (defaults to `True`).
- Maintains it in the `entityUpdate` match block: `case "gridReverse": self.exports_bypass = value != 2` (where `2 = forbidden` per the select definition in `entity.py:153`).
- The manager line is now a single readable expression: `self.discharge_bypass -= d.pwr_produced if d.state == DeviceState.SOCFULL and d.exports_bypass else 0`.

No more `entities.get()` / `isinstance` / mixed-type tuple in the hot loop, and the attribute is available for #1318 / #1330 to depend on without each reinventing the lookup.

Merging now. Thanks for the fix!

brain169:
There are little bugs in this fix (with device.py) which the previous did not have.

This happens with more than one battery unit. For example, if the household grid requires 200 watts while both batteries are currently depleted and the solar panels are supplying 100 watts each, this *should* actually cover the demand; however, the system draws only 100 watts from one battery, while the other remains at zero output. This state seems to persist until a large amount of power is suddenly required; at that point, there is a brief delay before the demand is once again met by drawing power 50/50 from both batteries.
Bildschirmfoto 2026-06-29 um 10 48 41
It somehow desperately tries to use only one memory module up to a certain demand level, whereas the previous version actually always distributed the load nicely.

FireSon:
Hi @brain169, thanks for the detailed report and the screenshot.

I traced through the scenario in the current code (master + 1.4.0-pre1):

- The battery with `homeOutput == 0` (its solar going to its own battery instead of to home) lands in `self.idle`, not `self.discharge`.
- The idle-device wake-up in `power_discharge` only fires when `d.electricLevel + 3 < self.idle_lvlmax`. When both batteries have similar low SOC, that condition stays false, so the idle one never gets commanded. A demand surge eventually pushes the device's firmware to start outputting on its own, and at that point the next cycle classifies it as discharge so the 50/50 distribution resumes.

The code paths involved here haven't changed in #1296 or any of the recent merges, so I don't think this is a regression introduced by that PR. To be sure though, could you confirm:

1. Which prior version you're comparing against (1.3.1, or older)?
2. Does the behavior still occur on 1.4.0-pre1?
3. A log snippet showing the `P1 ======>` line and the per-device state when one battery is stuck at 0 W would help a lot.

If it still occurs on 1.4.0-pre1, the fix is in the wake-up heuristic. That's a separate PR, not a release blocker.

Thanks again.

brain169:
> Hi @brain169, thanks for the detailed report and the screenshot.
> 
> I traced through the scenario in the current code (master + 1.4.0-pre1):
> 
> * The battery with `homeOutput == 0` (its solar going to its own battery instead of to home) lands in `self.idle`, not `self.discharge`.
> * The idle-device wake-up in `power_discharge` only fires when `d.electricLevel + 3 < self.idle_lvlmax`. When both batteries have similar low SOC, that condition stays false, so the idle one never gets commanded. A demand surge eventually pushes the device's firmware to start outputting on its own, and at that point the next cycle classifies it as discharge so the 50/50 distribution resumes.
> 
> The code paths involved here haven't changed in #1296 or any of the recent merges, so I don't think this is a regression introduced by that PR. To be sure though, could you confirm:
> 
> 1. Which prior version you're comparing against (1.3.1, or older)?
> 2. Does the behavior still occur on 1.4.0-pre1?
> 3. A log snippet showing the `P1 ======>` line and the per-device state when one battery is stuck at 0 W would help a lot.
> 
> If it still occurs on 1.4.0-pre1, the fix is in the wake-up heuristic. That's a separate PR, not a release blocker.
> 
> Thanks again.

,Thanks for the detailed reply! If nothing has changed regarding the fix in that specific area, it’s likely a different combination of factors. Unfortunately, it’s hard to observe, and I can’t say with 100% certainty whether this happened before version 1.3.1. I’ve since installed the Zendure 2.0 update, too. The problem doesn't seem to occur every day, either. If I do end up installing the pre-release version, I’ll let you know.

Issue #1230, opened 2026-03-25
[Bug] SolarFlow AC/AC+ batteries failing to start charging due to mathematical sign conflict in power distribution logic


thib3113:
### Home Assistant Version
2026.3.4

### Zendure Integration Version
1.2.7-pre3

### Describe the bug
I have observed a bug where my SolarFlow batteries (**sf2400ac** and **sf2400ac+**) remain in `INACTIVE` state even when there is a significant solar surplus (over 1000W of grid injection). The batteries stay in "Idle" mode and refuse to transition to "Charge" mode because the internal power distribution logic fails to handle the negative sign convention used by these AC models.

### To Reproduce
1. Have a SolarFlow AC or AC+ battery in `INACTIVE` state (Idle).
2. Configure the device with a fusegroup (in my case: **"Device has its own circuit or phase"** for each battery).
3. Generate a solar surplus (negative P1 value) between 0W and 1200W.
3. Observe that the battery remains in `INACTIVE` state and does not start charging, despite the surplus being available.
4. The system only recovers if the surplus exceeds ~1200W or after a long delay via the "50W pulse" fallback mechanism.

### Expected Behaviour
The batteries should immediately transition to `CHARGE` mode as soon as a negative surplus (injection) is detected on the P1 meter, respecting the `SmartMode.POWER_START` threshold.

### What device are you using?
- SF2400 AC
- SF2400 AC+

### Log Extracts
```shell
2026-03-25 09:13:50 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-1069 isFast:False, setpoint:-1069W stored:0W
2026-03-25 09:13:50 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -1069W
2026-03-25 09:13:50 DEBUG (MainThread) [custom_components.zendure_ha.manager] Device sf2400ac 063 state: INACTIVE, SOC: 12%, homeInput: 0W, homeOutput: 0W
2026-03-25 09:13:50 DEBUG (MainThread) [custom_components.zendure_ha.manager] Adding sf2400ac 063 to idle list
```

#### The "HTTP Timeout Rescue" (Strange Recovery)
Ironically, in my case, the only thing that eventually "saves" the system and starts the charge is a communication bottleneck on the battery side. The internal ESP processor sometimes fails to keep up with the volume of requests, leading to timeouts.
1. The system is stuck in the 0W bootstrap loop.
[...]

PR #1297, opened 2026-04-28
Correct power start value for solarflow 2400 AC+

zoic21:
This change updates the minimum power required to start charging from 50W to 60W for the SolarFlow 2400 AC+.

The previous threshold could fail to trigger charging.

This adjustment may also be applicable to the SolarFlow 2400 AC, but has only been validated on the AC+ model.

### DIFF
--- a/custom_components/zendure_ha/const.py
+++ b/custom_components/zendure_ha/const.py
@@ -69,5 +69,5 @@ class SmartMode:
 
     HEMSOFF_TIMEOUT = 60  # Seconds before HEMS state is set to OFF if no updates are received
 
-    POWER_START = 50  # Minimum Power (W) for starting a device
+    POWER_START = 60  # Minimum Power (W) for starting a device
     POWER_TOLERANCE = 5  # Device-level power tolerance (W) before updating


cybe:
This should have been already addressed by https://github.com/Zendure/Zendure-HA/commit/f787a6a38c3bd23e07bcad7c7e64cd3859b9579e. Maybe I'm wrong though.

thib3113:
Also, 60w is not always enough ... even with 100w my 2400 AC+ doesn't start correctly each times ... ( in input or in output ) ...

zoic21:
> This should have been already addressed by [f787a6a](https://github.com/Zendure/Zendure-HA/commit/f787a6a38c3bd23e07bcad7c7e64cd3859b9579e). Maybe I'm wrong though.

I try in 1.3.0 with 50 device never start charging, I change by 60 restart HA device start charging without issue.

zoic21:
> Also, 60w is not always enough ... even with 100w my 2400 AC+ doesn't start correctly each times ... ( in input or in output ) ...

How many AB3000L you have ? Maybe is the min power start is link to that, on my side I don't have any AB3000L for now.

thib3113:
> How many AB3000L you have ? Maybe is the min power start is link to that, on my side I don't have any AB3000L for now.

One, but I already opened an issue here : https://github.com/Zendure/Zendure-HA/issues/1230

an0Nym0us63:
If it was an issue for the 2400 AC+ it should have been a value dedicated for it

I have  2 2400 AC with 3  AB3000 each and i never had any issue during more than 6 months with 50W. Why should people be kind of penalized for one specific device that would need tweakings

thib3113:
> Why should people be kind of penalized for one specific device that would need tweakings

@an0Nym0us63 it just a "start value", when the battery start to output, the value will be changed . 

Not sure to understand the problem ? ( it just a command to wake up the battery, not a minimum )

an0Nym0us63:
> > Why should people be kind of penalized for one specific device that would need tweakings
> 
> @an0Nym0us63 it just a "start value", when the battery start to output, the value will be changed .
> 
> Not sure to understand the problem ? ( it just a command to wake up the battery, not a minimum )

AS i understand the code its the minimum value of export needed to trigger the start of the charging process.


In case of cloudy day or if you have lots of loads at house (A/C end so on)  that will make 10w to wait each cycle (of positive import/ positive export) before restarting the charging process



I may be wrong....

thib3113:
@an0Nym0us63 

Actually, taking another look at the code, it seems the POWER_START variable serves two different purposes!

You were totally spot on about the cloudy day scenario. In the power distribution logic, it acts as a strict minimum threshold to trigger the charging process. If the solar surplus doesn't reach that 60W mark, the condition isn't met, the device isn't selected, and the battery stays idle.

However, it also does what I initially thought: it is used as the baseline value for the 'wake-up' command sent to start the battery. The code even applies a specific tweak to this value (subtracting 10W) to force the hardware to wake up when it's stuck.

an0Nym0us63:
@thib3113 Yes we were both right and my comment was about the power distribution logic. And on my point of view it is kind of a bad idea to penalize the global system for only some devices that needs tweakings.

zoic21:
I am not sure it’s really penalize all other device but I Will see to make this parameters by device.

an0Nym0us63:
> I am not sure it’s really penalize all other device but I Will see to make this parameters by device.

Hey @zoic21  how are you doing ?  Do you remember me ?

zoic21:
> > I am not sure it’s really penalize all other device but I Will see to make this parameters by device.
> 
> Hey @zoic21 how are you doing ? Do you remember me ?

Oooooooooo yes how are you ? it's fun to see you here but home automation is small world.

an0Nym0us63:
> > > I am not sure it’s really penalize all other device but I Will see to make this parameters by device.
> > 
> > 
> > Hey @zoic21 how are you doing ? Do you remember me ?
> 
> Oooooooooo yes how are you ? it's fun to see you here but home automation is small world.

Very fine ;) Of course it's a small world, it would be great to discuss "le bout de gras" one day to see how the things went since i left ....dom

PR #1318, opened 2026-04-29
Fix discharge_bypass interactions with non-SOCFULL devices

ahocevar:
Fixes #1302. Probably also fixes #1312, depending on the reporter's setup — same symptom shape. #1304 may sound similar, but is definitely not fixed by this.

## Background

#1151 reported SOCFULL Hypers (bypass forbidden) cycling output to 0W: with the original `setpoint -= self.discharge_bypass`, setpoint could flip sign when home demand was below the SOCFULL bypass solar, sending the manager into `power_charge()` against devices that physically couldn't accept charge.

#1162 fixed it by clamping the result at 0 when `p1 ≥ 0`. This works for the #1151 topology (SOCFULL devices, no other absorber/discharger), but is too coarse for the rest of the configuration space — the clamp now runs whether or not other non-SOCFULL devices are present, and that introduces two new failure modes.

## Regressions introduced by #1162

**Charge-side oscillation.** With SOCFULL Hypers passing solar and a non-SOCFULL absorber (e.g. SolarFlow 2400 AC) trying to take the surplus, the absorber's draw flips P1 slightly positive. The clamp's `max(0, …)` then zeroes setpoint — the manager calls `power_discharge(0)`, which tears down the in-flight charge. ~60–90s cycle, never sustains charging.

<details>
<summary>Log excerpt: charge-side oscillation (one cycle, ~100s)</summary>

```
12:02:10  P1 ======> p1:-403  isFast:False  setpoint:-403W
12:02:10  Charge => setpoint -403W
12:02:10  Power charge SolarFlow 2400 AC => -60     ← ramp starts
12:02:26  P1 ======> p1:-339  isFast:False  setpoint:-398W
12:02:26  Power charge SolarFlow 2400 AC => -398    ← full charge
12:02:33  P1 ======> p1:-115  isFast:True   setpoint:-431W
12:02:33  Power charge SolarFlow 2400 AC => -431
12:02:43  P1 ======> p1:+42   isFast:False  setpoint:0W      ← clamp forces 0
12:02:43  Discharge => setpoint 0W
12:02:43  Power discharge SolarFlow 2400 AC => 0    ← charge torn down
12:02:52  Charge => setpoint -360W
12:02:52  Power charge SolarFlow 2400 AC => 0       ← hysteresis lockout
12:02:59  Charge => setpoint -406W                  ← no power command
12:03:13  Charge => setpoint -405W                  ← no power command
12:03:27  Charge => setpoint -407W                  ← no power command
[...]

PR #1325, opened 2026-04-30
Avoid grid round-trip when charge target can't overcome off-grid pull

ahocevar:
When a device with an active off-grid AC outlet is in `acMode=1` (charge), the device sources **both** battery charging and the off-grid load from `gridInputPower`, so `homeInput ≈ |pwr| + pwr_offgrid`. If the assigned charge target `|pwr|` is smaller than `pwr_offgrid`, sending the charge command makes the system import `pwr_offgrid + |pwr|` to absorb a `|pwr|`-sized surplus — net grid import goes *up*, not down.

This patch detects that regime in `power_charge()` and swaps the device to `acMode=2 outputLimit=10` instead, so the battery supplies its own off-grid load and the small surplus feeds home directly. Large surpluses (`|pwr| > pwr_offgrid`) still charge normally — the swap is gated on `pwr > -d.pwr_offgrid`.

## Reproducing

Two Hypers + SolarFlow 2400 AC with an off-grid load (~30W). With Hyper feed-to-grid set to `forbidden`, the autonomous solar feed stops, exposing the SF's off-grid pull on the P1 meter. System steady-state observed:
```
P1 ======> p1:45 isFast:False, setpoint:-14W stored:25W
Charge => setpoint -14W
Power charge SolarFlow 2400 AC => -14
```

`setpoint − p1 = -59` ⇒ `SF.homeInput ≈ 59W = ~14W battery charge + ~45W gridOffPower`. The system commands -14W of battery charge, which drags 45W of off-grid pull along with it — `p1` settles at ≈`gridOffPower` instead of 0.

## When the swap fires

The swap only fires when the assigned charge target is smaller than the device's off-grid pull (`|pwr| < pwr_offgrid`). In that regime — typically when no upstream device is supplying home demand, e.g. when feed-to-grid is forbidden — `acMode=1` would source the off-grid leg from grid alongside the battery charge, so charging amplifies grid import rather than absorbing surplus. Outside that regime (`|pwr| ≥ pwr_offgrid`, e.g. during meaningful solar absorption), the device charges normally.

## Scope of behavior change

- Per-device check inside the charge distribution loop, gated on `d.pwr_offgrid > 0 AND pwr > -d.pwr_offgrid`.
- Large solar absorption is untouched: setpoint of -1262W with `pwr_offgrid=45W` gives `pwr=-1262`, `-1262 > -45` is false, normal charge path runs.
- Devices without an off-grid outlet (`pwr_offgrid == 0`) are unaffected — `pwr_offgrid > 0` short-circuits the check.
- Existing precedent for `power_discharge(10)` on off-grid charge devices: same value, same intent (avoid reclassification as charging on next refresh).

## Result

Same test scenario, after fix: `p1` settles to ≈ `-house_solar` (small feed-in) instead of `+pwr_offgrid` consumption — closer to zero in absolute terms, and no longer paying for the off-grid load through the grid meter.

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -550,6 +550,17 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                 self.pwr_low = 0 if (delta := d.charge_start * 1.5 - pwr) >= 0 else self.pwr_low + int(-delta)
                 pwr = 0 if self.pwr_low < d.charge_optimal else pwr
 
+            # In acMode=1 the off-grid outlet is sourced from grid alongside
+            # battery charge, so homeInput ≈ |pwr| + pwr_offgrid. If the share
+            # we'd charge doesn't exceed the offgrid pull, the offgrid leg
+            # dominates and the surplus we wanted to absorb turns into net
+            # import. Switch to acMode=2 so battery supplies offgrid; the
+            # outputLimit=10 (vs 0) keeps the device in the discharge bucket
+            # next refresh instead of being reclassified as charging.
+            if d.pwr_offgrid > 0 and pwr > -d.pwr_offgrid:
+                await d.power_discharge(10)
+                continue
+
             setpoint -= await d.power_charge(pwr)
             dev_start += -1 if pwr != 0 and d.electricLevel.asInt > self.idle_lvlmin + 3 else 0
 


fellek:
Hi @ahocevar thanks for your analysis. I had other issues with offgrid before (~110W) and it would be interesting, if you have tested your fix with offgrid power higher than 100W? 
SF2400AC does not show selfconsumption as an own entity. I suggest for my SF2400AC selfconsumption for offgrid-socket around ~10-14W and on charging 20-40W. I think thats why `POWER_START` (now 60W, former 50W) is a reasonable value for start charging, but also a limit, when stop charging, if surplus power is lower than `POWER_START`.

In that case, it could be like that:  `d.pwr_offgrid > 0 AND pwr > POWER_START`
In other Words: "surplus has to be > POWER_START + consumed offgrid power" for charging. otherwise do not charge. 
Do you confirm this?

ahocevar:
@fellek Yes, I confirm. Your reasoning is sound and aligns with the POWER_START hysteresis used elsewhere in `manager.py` (around lines 525 and 596). My current guard only catches the strict sign-flip case (`pwr + pwr_offgrid > 0`), which leaves a band where the net charge share exceeds the off-grid pull by only a few watts — well within self-consumption (your 20–40W on charge) and round-trip loss territory. In that band we'd still see net import even though the inequality says "charging wins."

Tightening to your proposal:
```python
if d.pwr_offgrid > 0 and pwr + d.pwr_offgrid > -SmartMode.POWER_START:
    await d.power_discharge(10)
    continue
```
…makes the threshold "surplus must exceed POWER_START + offgrid pull before we commit to charging," which matches the start/stop semantics POWER_START already has and gives the same margin the rest of the charge-path uses.

On testing with offgrid > 100W: I haven't deliberately exercised the 110W+ regime. My repro was in the 30–80W band. Given the linearity of the relation (`homeInput ≈ |pwr| + pwr_offgrid`), the fix should scale, but I take your point that the previous 110W incident is a known sore spot. Happy to update the condition to use POWER_START as you suggest before merging. That also gives a wider safety margin if SF2400AC self-consumption is higher than I assumed. Want me to push that change?

fellek:
Hi @ahocevar, I will test your code with offgrid= ~140W from now on and log the sunrise-effects, when surplus will rise slowly.

ahocevar:
I'm not so sure about this pull request any more. It breaks again what I fixed in #1318. I need to think more about it, maybe I can come up with a better fix. But I'm closing this for now.

fellek:
thanks for info!

PR #1330, opened 2026-05-03
Don't suppress discharge for gridReverse=forbidden devices

ahocevar:
The manager's idle-startup buffer (`discharge_optimal*2`) and optimal-range hysteresis both assume small unmet loads can flow to grid harmlessly, so they keep devices at 0 W rather than have them run at sub-optimal low
power. With `gridReverse=forbidden` the trade-off inverts — every unmet watt becomes a paid import — so skip both per-device for forbidden devices.

This pull request complements #1296, which fixes the same `gridReverse=forbidden` symptom at the bucketing/bypass step.

Repro: 2 SOCFULL Hyper 2000 + ~80 W home load, all devices set to forbidden. Before: idle-buffer keeps the Hypers parked, 80 W imported. After: a forbidden device wakes and covers the load, P1 settles near 0.

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -581,8 +581,16 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
             # SF 2400 may show more gridInputPower than offGridPower and will be recognized as charging, so set power to 10 instead of 0
             await d.power_discharge(0 if max(0, d.pwr_offgrid) == 0 else 10)
 
-        # distribute discharging devices, use produced power first, before adding another device
-        dev_start = max(0, setpoint - self.discharge_optimal * 2 - self.discharge_produced) if setpoint > SmartMode.POWER_START else 0
+        # distribute discharging devices, use produced power first, before adding another device.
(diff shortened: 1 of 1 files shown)


FireSon:
Just a copy of my other comment...
@ahocevar  Thanks for the PR! At this moment there are 3 different PR's handling gridReverse=forbidden. They all solve different parts if I understand it correctly. 
#1296 (zaubara) — fixes how discharge_bypass is accumulated (don't accumulate for forbidden devices)
#1318 (ahocevar) — fixes how discharge_bypass is consumed in the setpoint calculation
#1330 (ahocevar) — fixes how forbidden devices affect dev_start and pwr_low hysteresis in power_discharge

I would like to keep things as generic as possible, so no SF2400 specials in the manager. But maybe I am too quick to judge since the cause and effect is not totally clear to me. (I do not have solar panels connected to my SF2400) Can you create a small functional design explaining why and what and where a change is necessary? 
That would be great!
Thanks

ahocevar:
@FireSon All of the mentioned pull requests are obsolete. I'm closing this one, #1318 is closed already. #1298 should be closed too.

PR #1331, opened 2026-05-03
Smooth P1 with rolling average to dampen control overshoot

ahocevar:
The manager re-issues device commands on every P1 sample, but device output follows with ~1-2 s of firmware lag (Hyper 2000 in my case). Reacting to raw P1 makes each correction overshoot into the next cycle's reading, and the loop never settles — observed as ±50 W of steady-state jitter.

This pull request makes it so we use the mean of the rolling 8-sample p1_history as the setpoint input. This trades ~20 s of phase lag for roughly 3× less amplitude. The fast-update path is unaffected: when isFast fires, p1_history is cleared and the mean equals the new sample, so genuine load steps still get an immediate response.

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -408,7 +408,12 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                 self.produced = 0
                 for fg in self.fuseGroups:
                     fg.initPower = True
-                await self.powerChanged(p1, isFast, time)
+                # Pass an averaged p1 to dampen proportional overshoot from the
+                # control loop's per-cycle re-distribution. On a fast change the
+                # history was just cleared and holds only the current sample, so
+                # the average equals p1 — fast updates stay responsive.
+                p1_smooth = int(sum(self.p1_history) / len(self.p1_history))
+                await self.powerChanged(p1_smooth, isFast, time)
             except Exception as err:
                 _LOGGER.error(err)
                 _LOGGER.error(traceback.format_exc())


harrymayr:
The default Normal update interval (TIMEZERO) is 4 seconds, so I don't see how this overshoot in the next cycle, assuming a time lag of 1-2 seconds.

ahocevar:
Thanks for your review, @harrymayr! Re-reading my commit message and your comment, the "overshoot" framing was just wrong. Yes, the smoothing only takes effect on the normal path (isFast clears the deque so smoothed equals raw), and on the normal path your math holds: 4s + 1-2s lag = settled before the next cycle. There's no overshoot in the only path the fix touches.

What the fix actually does: smooth P1 on the normal path so the manager's per-cycle setpoint is more stable across consecutive 4-15s cycles. The 5W POWER_TOLERANCE on each device is a command-resolution threshold (clearly intentional and works fine at that level). What seems missing is a setpoint-level deadband — a check at the manager level that doesn't re-bucket/re-distribute when the inputs haven't changed by some threshold from the last cycle. The averaging this pull request suggests works around that absence by smoothing the input rather than gating the recomputation, but it's solving the same noise-propagation problem upstream.

If you'd prefer the more principled fix (an actual setpoint deadband at the start of `powerChanged`), I'm happy to close this pull request and open one for that instead. I considered the averaging change smaller and lower-risk, but you're right if this feels like a workaround rather than a fix to you.

For now, I at least amended the commit message to better describe what this fix actually does.

FireSon:
@ahocevar Thanks (again 😄) for another PR, it correctly replaces p1 with p1_smooth in the powerChanged call. The open question is deadband vs smoothing, which is a design preference. So I am not sure why this needs to be changed. I am doing some test in a new version of the integration, and from my tests, I found the less averaging you do the better it becomes. It is fair to say that in that version I also take lag of the devices into consideration, so it is not a one to one comparison.

ahocevar:
@FireSon You convinced me re averaging vs deadband. Looking at my changes here now, and given the small benefit my system gained from it in practice, I'm closing this. If the observed oscillation bothers me again in the future, I'll consider a deadband approach instead, but I think for now we're good.

PR #1334, opened 2026-05-03
Force-charge ACE 1500 in UPS mode below reserve target

ahocevar:
When an ACE 1500 is set up as a UPS — AC input connected, AC output serving the load — the manager has no way to top up the battery from grid when there's no surplus. Reserve drifts to minSoc and stays there, defeating the point of UPS mode (a power outage finds the battery empty).

This pull request add a `ups_mode` property on the base ZendureDevice (default False) and overrides it on ACE1500 to detect `packState == 3`, the firmware-level signal that AC input is available and the device is running passthrough. After the regular discharge dispatch, we now scan for UPS-mode devices below the midpoint of `[minSoc, socSet]` and command them to charge at POWER_START (60 W) from grid. The pass runs at the end of `power_discharge` so it overrides the `power_discharge(0)` the top loop sends to charge-bucket members.

Reserve target = midpoint keeps half the usable capacity for outage backup without holding the battery near 100 %. POWER_START as the charge rate is gentle on the device fans (no nighttime fan noise) and modest on grid use even with multiple ACEs topping up at once. Stateless: no Schmitt trigger between `minSoc` and `target` — small self-consumption drift re-triggers a small top-up rather than letting SoC fall all the way to minSoc first.

This pull request is related to #1256, which discusses ACE 1500 UPS scenarios. It addresses the "keep the battery reserve charged from grid" half of the workflow @Popoboxxo described; the discharge-regulation half remains open.

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -644,6 +644,13 @@ class ZendureDevice(EntityDevice):
         """Get the offgrid power."""
         return 0
 
+    @property
+    def ups_mode(self) -> bool:
+        """True when the device is acting as a UPS (grid passthrough with
+        battery backup). The manager uses this to maintain a charge reserve
+        from grid when no surplus is available."""
+        return False
+
 
 class ZendureLegacy(ZendureDevice):
     """Zendure Legacy class for devices."""
--- a/custom_components/zendure_ha/devices/ace1500.py
+++ b/custom_components/zendure_ha/devices/ace1500.py
@@ -21,6 +21,15 @@ class ACE1500(ZendureLegacy):
         self.acSwitch = ZendureSwitch(self, "acSwitch", self.entityWrite, None, "switch",1)
         self.dcSwitch = ZendureSelect(self, "dcSwitch", {0: "off", 1: "on"}, self.entityWrite, 1)
 
+    @property
+    def ups_mode(self) -> bool:
+        """ACE 1500 reports packState=3 when AC input is available — i.e. the
+        AC load runs as grid passthrough with the battery on standby for an
+        outage. packState falls to 2 (discharging) when AC is disconnected,
+        and to 0 (sleeping) when DC input takes over."""
+        pack_state = self.entities.get("packState")
+        return pack_state is not None and pack_state.asInt == 3
+
     async def charge(self, power: int) -> int:
         _LOGGER.info("Power charge %s => %s", self.name, power)
         self.mqttInvoke(
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -628,3 +628,15 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                     if (dev_start := dev_start - d.discharge_optimal * 2) <= 0:
                         break
             self.pwr_low: int = 0
+
+        # Force-charge UPS-mode devices that have drifted below their reserve
+        # target. Target is the midpoint between minSoc and socSet, keeping
+        # half the usable capacity in reserve for a grid outage. Runs after
+        # the regular dispatch so this command wins over the power_discharge(0)
+        # the top loop sent.
+        for d in self.devices:
+            if not d.online or not d.ups_mode:
+                continue
+            target = (d.minSoc.asNumber + d.socSet.asNumber) / 2
+            if d.electricLevel.asNumber < target:
+                await d.power_charge(-SmartMode.POWER_START)
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -299,6 +299,20 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         self.update_count += 1
         self.totalKwh.update_value(kwh)
 
+        # Force-charge UPS-mode devices that have drifted below their reserve
+        # target. Target is the midpoint between minSoc and socSet, keeping
+        # half the usable capacity in reserve for a grid outage. Runs every
+        # SCAN_INTERVAL so it covers all manager modes (OFF, MATCHING with
+        # surplus, MATCHING_CHARGE, STORE_SOLAR, MANUAL with negative power)
+        # — UPS reserve is a hard constraint independent of how the manager
+        # is otherwise dispatching power.
+        for device in self.devices:
+            if not device.online or not device.ups_mode:
+                continue
+            target = (device.minSoc.asNumber + device.socSet.asNumber) / 2
+            if device.electricLevel.asNumber < target:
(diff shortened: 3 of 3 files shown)


fellek:
@ahocevar I like your idea to introduce an UPS-Mode. 

Using POWER_START as recharge-power e.g. after grid-loss for 1hour, to desired soc-level would be a very inefficent way, because your device would have ~40W or more selfconsumption and ~20W (or less) chargingpower (+ offgrid-consumption) from minSoc to desired SOC would take a long time. 
What do you think about recharging battery with more power to keep charging time short and keep UPS prepared with desired capcity for next blackout? That would increase efficiency, but also noise, if there will be heat that needs to be cooled.

Does that match your idea for UPS-Mode?

ahocevar:
@fellek The problem is that in UPS mode, the ACE's self consumption is supplied from battery, so the battery is slowly draining. After a grid loss, I agree it would be desirable to charge as quickly as possible. But since the charging threshold can be reached any time even when the grid is operating normally, I'd prefer to avoid the noise. I noticed that even at 100W charging power, there is noise that I don't want at night.

I think detecting a grid loss and handling it with a separate code path with fast charging upon grid recovery would be possible, but a bigger effort — probably worth a separate pull request. My goal for this pull request was to prevent the battery from fully draining under normal grid operation.

fellek:
I understand, this is a relevant limitation for your system. just for my understanding: in UPS-mode, offgrid socket has stable power delivery, but drains low power from Battery for selfconsumption. In which state will be the device in case of charging? A) will it stay in real UPS-mode during charge, or B) will UPS-function will be less reliable at this time?

ahocevar:
@fellek Correct. In UPS mode (not an invention of this pull request, but a firmware feature of the ACE 1500), the off-grid socket provides stable power delivery. When no AC charging is configured, the load will be supplied from the grid, but the ACE's self consumption will be drawn from battery. When AC charging is configured, the battery will be kept at maxSoc. In both cases (charging or not charging), there is stable supply for the off-grid load.

Setting AC charging to a fixed power (or appointment or cost based — only available in the Zendure app) has the disadvantage of leaving no room for surplus charge from solar (see #1338 for a fix that fixes ACE 1500 charging from solar surplus).

fellek:
ok, let me understand last point: 
is this calculation possible? `inputLimit=selfconsumption + Offgrid + zerocharging`
what will happens to offgrid-power, if the device looses grid in this calculation? will it stay stable?

ahocevar:
That calculation is done by the ACE firmware. Off-grid supply will always be stable no matter what you configure as charging, until MinSoc is reached and no AC grid is available. Then the ACE will turn off off-grid supply.

Even if you set charging to 0, the off-grid outlet will continue to work when the SOC is at minSoc.

ahocevar:
Maybe a summary for better understanding: the problem that this pull request solves is self-consumption discharge of ACE 1500 batteries when in UPS mode (AC input and AC off-grid output active) with AC charging deactivated.

Why would someone configure an ACE 1500 this way? To leave battery charge capacity for solar power (either connected to ACE 1500's inverter or surplus from the Zendure manger, see #1338)

If no DC charge source (e.g. solar input) provides any charge, the batteries will drain until they reach minSOC, leaving no power for UPS in case of a grid failure.

By letting the Zendure manager set a small AC input when the SOC drops to somewhere above minSOC but below half of the available charge, we make sure that even in the absence of another charge source, there will be some power left for the AC failure case.

harrymayr:
You can't start Startpower AC charging in UPS-Mode, because there is already some AC-load. The only way I see is the charge the ACE 1500 with the **offGridpower + self discharge power** once the desired reserve socLimit is reached. And this is necessary not only in the discharge loop. I think this should be independend from the manager modes.

ahocevar:
Thanks for the careful review, @harrymayr. Two responses:

> You can't start Startpower AC charging in UPS-Mode, because there is already some AC-load.

I had the same concern initially, but testing shows the opposite: in UPS mode, the ACE 1500 draws exactly `pwr_offgrid` from grid for the AC passthrough load. When AC charging is additionally commanded, the device draws `pwr_offgrid + chargingPower` from grid — the two coexist fine. There's no firmware refusal.

What actually motivates this pull request is a different problem: when *no* AC charging is commanded, the device powers its own internal electronics from the battery rather than from the AC input. SoC slowly drains. The force-charge logic counters that drain so the UPS reserve doesn't quietly disappear.

> And this is necessary not only in the discharge loop. I think this should be independent from the manager modes.

Agreed. I've moved the UPS pass from the end of `power_discharge` to `_async_update_data` so it fires every SCAN_INTERVAL regardless of manager mode. That covers OFF, MATCHING with surplus (-> power_charge), MATCHING_CHARGE, STORE_SOLAR, and MANUAL with negative power, all of which previously skipped the check.

harrymayr:
> I had the same concern initially, but testing shows the opposite: in UPS mode, the ACE 1500 draws exactly `pwr_offgrid` from grid for the AC passthrough load. When AC charging is additionally commanded, the device draws `pwr_offgrid + chargingPower` from grid — the two coexist fine. There's no firmware refusal.

OK, this is completely different to i.e. the SF 2400 AC. With offGrid-Load and setting `charge(100)` it will use 100W from the grid (gridInputPower), and the remaining offGrid power needs from the battery.

However, this is something which need to be taken into account for the distribution and the handling of the offGridPower.

ahocevar:
> However, this is something which need to be taken into account for the distribution and the handling of the offGridPower.

The ACE 1500 is a lot different from the SF 2400 AC in this regard as well: In UPS mode, the offGridPower is just a passthrough from AC (grid), so it should already be accounted for.

harrymayr:
Can you confirm, that the `inputLimit` is always the power which correspond to the `ouputPackPower` (power into the battery) independend of the offGridPower/`acOutputPower`? So that the following formula is true:
`gridInputPower = acOutputPower (+ dcOutputPower) + inputLimit (or outputPackPower)`
Perhaps `outputPackPower` is not always `inputLimit` because of the power which is needed for the device itself (`batterieElectric`?)

ahocevar:
@harrymayr I can confirm that `outputPackPower` never exceeds `inputLimit`. Of course it can be lower when the battery is almost full, and `0` when maxSOC is reached. The ACE seems to draw the power for its internals from the battery, but that's not shown by any of the sensors except for the slow decrease in battery SOC that can be observed.

FireSon:
@ahocevar Why do we need a separate UPS mode? If the device can charge, why not just charge it like all other batteries? It is possible to use solar power from devices to charge others. I just want to understand the use case, which is a bit harder for me, since I do not own such a device.

ahocevar:
Closing because it is probably too specific to my use case. In my setup, I have now set up an automation to handle this: when battery SOC is below a threshold, switch to charging with a specified power until we're above another threshold.

PR #1338, opened 2026-05-04
Make ACE 1500 controllable without a Hub

ahocevar:
This pull request adds a `hubMode` toggle to the ACE 1500 device: "paired" (default,
preserves existing behavior) or "standalone" (ACE 1500 not connected to a Hub 1200/2000).

In paired mode, the integration sends Smart Matching commands (autoModel=8) which expect a Hub heartbeat to act on the chargingPower field — historical behavior, kept as default for backward compatibility.

In standalone mode the integration parks the device in autoModel=0 (None program) and drives charge/discharge through inputLimit / outputLimit property writes, the only path the ACE 1500 honors without a Hub.

Standalone mode is required for ACE 1500 owners without a Hub (which is not an uncommon configuration — the ACE 1500 has its own inverter and does not need to dock with Hub 1200/2000).

This pull request fixes the standalone-control half of #1256, addressing the observation by @harrymayr that "the only possibility to control AC charging is with the inputLimit". It is a prerequisite for #1334 to deliver value on standalone ACE 1500s.

### DIFF
--- a/custom_components/zendure_ha/devices/ace1500.py
+++ b/custom_components/zendure_ha/devices/ace1500.py
@@ -144,15 +144,18 @@ class ACE1500(ZendureLegacy):
         and rate-limiting. Each property write goes to the ACE 1500's flash
         memory, so an unguarded 5-second control loop would burn through
         endurance in months. Quantize to 50 W steps and throttle to one
-        write per 30 s. target=0 (stop charging) is exempted from the rate
-        limit so we can drop charging promptly when surplus disappears."""
+        write per 30 s — applies uniformly to start and stop so a surplus
+        that briefly crosses the quantization step doesn't trigger a
+        50W-write immediately followed by a 0W-write. The cost is that a
+        sustained drop in surplus keeps the device drawing its last
+        commanded value for up to one interval; that's bounded and small,
+        and worth it for flash endurance."""
         target = (target // _INPUT_LIMIT_STEP_W) * _INPUT_LIMIT_STEP_W
         now = datetime.now()
         if target == self._last_input_limit:
             return
         if (
-            target != 0
-            and self._last_input_limit is not None
+            self._last_input_limit is not None
             and now - self._last_input_limit_time < _INPUT_LIMIT_MIN_INTERVAL
         ):
             return
--- a/custom_components/zendure_ha/devices/ace1500.py
+++ b/custom_components/zendure_ha/devices/ace1500.py
@@ -1,6 +1,7 @@
 """Module for the ACE1500 device integration in Home Assistant."""
 
 import logging
+from datetime import datetime, timedelta
 from typing import Any
 
 from homeassistant.core import HomeAssistant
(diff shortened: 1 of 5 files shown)


harrymayr:
Correct me if I'm wrong (I don't have an ACE 1500) but according to the user manual I see no discharge possibility into the grid in standalone mode. I see only the possibility to discharge into the USB, XT-60 and offGrid-Sockets. The offGrid-Sockets and the USB are typically not power controllable, only switching ON and OFF. Maybe the XT-60 output is power controllable.
For me a discharge control in standalone mode doesn't make sense.
With respect to charge in standalone mode, you need to have in mind, that every input- or outputLimit change will be written direct into the flash-memory.
If you use the charging mode with the integration (i.e. with AC coupled PV), there are many changes of the charging-power depending on the sun and the home consumption.
Sooner or later the flash-memory will be damaged.
Coupled with the Hub, it's a little bit different.

So standalone charging is perhaps better in 50-100W steps and some delay before the next change of the power is possible.

image

ahocevar:
@harrymayr Thanks for your comments! You're absolutely right — my thinking regarding output limit was wrong. I changed the code as follows:
* No more output limit changes. The only thing left to do is set the input limit to 0 when we're not charging from surplus any more.
* Avoid repeated writes of the same values, quantize changes in input limit to 50 W steps, and throttle writes to one per 30s. I added constants for these so they can be changed easily. The only exception (`target=0` – stop charging) is exempted from the rate limit so we can drop charging promptly when surplus disappears.
* `_setAutoModelNone` runs only when a write actually happens — folded inside the dedup/throttle gate so we don't send redundant function/invoke calls either.

harrymayr:
2 more remarks:

- adding translation for french and dutch is useful (Google is your friend 😉)
- The limits need to be set according to the `hubMode` (and need to be updated if `hubMode` changes), so that the integration don't want to use the device for discharge in standalone mode. If there is a limit for discharge, the integration will distribute the needed power accross the other devices. This distribution then does not fit and the other devices will discharge too less.

ahocevar:
Ok, I've added fr and nl translations — totally trusting Google here. And good point on updating the limits on hub mode change. I added a commit which I think addresses your concern here. Let me know if I'm still missing something.

ahocevar:
Pushed df82b46 to drop the `target=0` exemption — observing 50→0 flash-write spikes when surplus oscillates around the 50W quantization step. Now the 30s interval applies uniformly: once we've just written any value, neither another non-zero target nor 0 can write again for 30s. Worst case is the device holding its last commanded value for up to one interval after a real surplus drop, which is bounded (≤0.4 Wh at 50W, ≤4 Wh at 500W) and seems worth it for flash endurance.

igosoft:
I'm going to test this week the changes in paired mode (ACE1500 with Hub2000)

ahocevar:
@github-actions Please reopen, this pull request would still be useful and has not received a follow-up review yet.

PR #1340, opened 2026-05-05
Engage producing devices for any home demand

ahocevar:
With the SF 800 Pro, I observed that it prefers to charge the battery until SOCFull before feeding the home demand, while home demand is fed from grid. With Zendure's HEMS, at least the off-grid demand is supplied by solar production.

This seems to be because the Zendure Manager has no path to engage the SF 800 once it's idle with production. The SF 800 is sitting at homeOutput = 0 and the manager is not commanding it:

* Bucketing condition `(home := d.homeOutput.asInt) > 0` is false.
* It lands in idle.
* Idle wake-up requires `dev_start = setpoint − discharge_optimal*2 − discharge_produced > 0`. With setpoint=3 and discharge_optimal*2 in the hundreds, that's 0.

So SF 800 stays unaddressed and keeps routing solar to its own battery. Whatever Zendure HEMS does, it apparently keeps the SF 800 actively engaged in output mode regardless of how small home demand is — likely by issuing setpoints continuously rather than only when home demand crosses some threshold.

What I successfully did to remedy this is change bucketing to treat `pwr_produced < 0` as "should be in discharge bucket".

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -436,8 +436,12 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                     self.charge_optimal += d.charge_optimal
                     self.charge_weight += d.pwr_max * (100 - d.electricLevel.asInt)
                     setpoint += -d.homeInput.asInt  # use gridInputPower directly; offgrid consumers are invisible to P1
-                # SOCEMPTY means, it could not discharge the battery, but it is still possible to feed into the home using solarpower or offGrid
-                elif (home := d.homeOutput.asInt) > 0:
+                # SOCEMPTY means, it could not discharge the battery, but it is still possible to feed into the home using solarpower or offGrid.
+                # Also include devices that are producing solar (pwr_produced < 0) but currently route it all to battery (homeOutput == 0).
+                # Without this, such devices land in the idle bucket where they are never engaged for small home demand, leaving solar to
+                # charge the device's own battery while home imports from grid. Bucketing them as discharge gives them a setpoint share so
+                # the manager can ramp homeOutput up; once homeOutput > 0 the original condition keeps them here on subsequent cycles.
+                elif (home := d.homeOutput.asInt) > 0 or d.pwr_produced < 0:
                     self.discharge.append(d)
                     self.discharge_bypass -= d.pwr_produced if d.state == DeviceState.SOCFULL else 0
                     self.discharge_limit += d.fuseGrp.discharge_limit(d)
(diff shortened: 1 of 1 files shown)


harrymayr:
I'm not sure, if this correct. Your PR set a device which get some solarInputPower always as discharge (without changing the setpoint).
The line `dev_start = max(0, setpoint - self.discharge_optimal * 2 - self.discharge_produced) if setpoint > SmartMode.POWER_START else 0` is intended to start an additional device, if there is a setpoint which is large enough to have the dischargepower in a usefull range for more devices. Assuming there is no device in discharge `- self.discharge_optimal * 2 - self.discharge_produced = 0` and `dev_start = setpoint` -> this start the first device with the highest SoC. The setpoint need to be larger than SmartMode.POWER_START.
If you have a second device and the setpoint is not large enough to start this device, it will be started (and the other stopped) once the SoC difference is >3%. This is the line `dev_start += 1 if pwr != 0 and d.electricLevel.asInt + 3 < self.idle_lvlmax else 0` 
If, in your case, the SF 800 pro has some solarInputPower but is idle, the SoC difference should be reached earlier and at some point the SF 800 should discharge, while the other device stops.

ahocevar:
The `dev_start` SoC-balancing path only runs when at least one device is already in the discharge bucket and receiving non-zero power (`pwr != 0` guard on line 624). When all devices are idle and setpoint ≤ `POWER_START`, `dev_start=0` and no idle device is ever started. The `pwr_produced < 0` condition puts the solar-producing device into the discharge bucket precisely so its production is counted in `discharge_produced` — the existing mechanism for tracking solar-available power — enabling it to cover small demands without battery use.

harrymayr:
But your solution count a device with solar power always as discharging. Assume a setpoint of 100W, a SF 2400 AC with SoC 90% and a SF 800 pro with SoC 50% and solarinputPower of 30W.
With your PR the 100W setpoint will the distributed on both devices, which is not very efficient.

ahocevar:
You're right — the weight-based distribution was assigning the SF 800 Pro a share of the setpoint that could exceed its solar output, forcing it to drain the battery.

I've amended the commit to add a cap: a device that entered the discharge bucket solely via pwr_produced < 0 (i.e. homeOutput is still 0) is limited to at most its current solar output (-d.pwr_produced). In your example the SF 800 Pro would receive at most 30 W, all from solar, while the SF 2400 AC covers the remaining 70 W from its battery (SoC 90%). Once the SF 800 Pro responds and homeOutput > 0, the cap is lifted and it participates in battery discharge normally.

FireSon:
@ahocevar Thanks for the PR! 
As a general rule of thumb, I do not like device specific 'fixes' in the manager. That should be generic code, and should be clean and as simple as possible. It seems to me that you are fixing a bypass issue, or do I miss something?

ahocevar:
When I apply #1352, I don‘t see the problem this pull request tries to fix anymore. So I‘m closing this.

PR #1341, opened 2026-05-05
Add global soc, delta cellule voltage and battery efficiency

zoic21:
Hello,

This PR introduces 3 new entities:

- **Global SOC** at the Zendure manager level
- **Battery voltage** delta at the battery level (maxVol - minVol)
- **Battery efficiency** aggrDischarge/aggrCharge*100

While both metrics could technically be derived using template sensors, exposing them natively improves usability and avoids unnecessary configuration on the user side.

**Use cases**
- Global SOC
  Provides a consolidated state of charge across the entire system, which is more relevant than per-battery values in most scenarios.
- Battery voltage delta
  Acts as a simple health indicator for individual batteries.
  A sustained delta greater than ~0.15V (e.g., over 2 hours) may indicate cell imbalance and suggest that the battery requires a full charge cycle for rebalancing.
- Battery efficiency
   Check efficiency of battery pack


custom_components/zendure_ha/device.py
```diff
@@ -85,6 +85,20 @@ def __init__(self, hass: HomeAssistant, sn: str, parent: EntityDevice) -> None:
...
+    def entityUpdate(self, key: Any, value: Any) -> bool:
+        """Update entity state and recalculate deltaVoltage when maxVol or minVol changes."""
+        changed = super().entityUpdate(key, value)
+        if changed and key in {"maxVol", "minVol"}:
+            max_vol = self.entities.get("maxVol")
+            min_vol = self.entities.get("minVol")
+            if max_vol is not None and min_vol is not None:
+                max_val = max_vol.asNumber
+                min_val = min_vol.asNumber
+                if max_val != 0 or min_val != 0:
```
harrymayr: I think AND is more usefull

zoic21: Hello,
Indeed, I change that.
Thanks.

FireSon:
Thanks @zoic21, nice contribution! The global SOC and delta-voltage are clean and clearly useful additions.

One thought on the efficiency part. The PR description describes it as:

> Battery efficiency `aggrDischarge/aggrCharge*100`

but the implementation does a bit more under the hood:

```python
raw_efficiency = (aggrDischarge + availableKwh) / aggrCharge
discharge_efficiency = sqrt(clamp(raw_efficiency, 0, 1))
efficiency = (aggrDischarge + availableKwh * discharge_efficiency) / aggrCharge * 100
```

I see what you're going for — crediting the still-stored energy with an assumed symmetric loss split is a clever idea. The concern is just that it makes the `efficiency` entity harder to reason about than the name suggests: a user looking at "Efficiency" will probably expect the straightforward round-trip ratio.

Would you be open to simplifying it to the plain formula from your description? Something like:

```python
self.efficiency.update_value(round(self.aggrDischarge.asNumber / charge * 100, 1) if (charge := self.aggrCharge.asNumber) > 0 else 0)
```

That matches what the entity name implies, and if the corrected version still feels valuable we could expose it as a second entity (e.g. `efficiency_corrected`) with its own description — best of both worlds.

Two tiny things while you're in there:
- `de.json` looks like it lost its trailing newline (`\ No newline at end of file` shows in the diff).
- A few translation entries have some stray whitespace inside the braces, e.g. `"global_soc": { ` — easy tidy-up.

Happy to merge once we've sorted the efficiency story. Thanks again for the work!

zoic21:
Hello,

Thanks for your review, indeed `aggrDischarge/aggrCharge*100` is more simple and pretty accuracy if you have lot of cycle.

For : 
- `de.json looks like it lost its trailing newline (\ No newline at end of file shows in the diff).`
 Don't understand why GitHub show that, if I check source file there is no return line after last }......
- `A few translation entries have some stray whitespace inside the braces, e.g. "global_soc": {  — easy tidy-up `
It's fix, and also some space at the end of translation

PR #1344, opened 2026-05-07
Fix overcurrent: correct power_charge tolerance check for ZenSdk devices

ahocevar:
This pull request fixes a bug affecting ZenSDK devices with an off-grid plug (e.g. SolarFlow 2400 AC): overcurrent on the off-grid plug due to a sign flip bug — the base `power_charge` tolerance check uses `power - homeInput`, but ZenSDK devices report `gridInputPower` as a *positive* value, so the check never converges and `charge()` is called on every manager tick. The resulting constant mode-switching triggered overcurrent events. Fixed by overriding `power_charge` in `ZendureZenSdk` to use `abs(power + homeInput)`.

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -735,6 +735,20 @@ class ZendureZenSdk(ZendureDevice):
 
         return await super().power_get()
 
+    async def power_charge(self, power: int) -> int:
+        """Set charge power for ZenSDK devices.
+
+        homeInput is reported as a positive value (grid consumption) for these
+        devices, so the inherited tolerance check (power - homeInput) would
+        never reach zero during charging.  Use power + homeInput instead,
+        which converges to zero when the device has settled at inputLimit=-power.
+        """
+        power = min(0, max(power, self.charge_limit))
+        if power < 0 and abs(power + self.homeInput.asInt) <= SmartMode.POWER_TOLERANCE:
+            _LOGGER.info("Power charge %s => no action [power %s]", self.name, power)
+            return power
+        return await self.charge(power)
+
     async def charge(self, power: int, _off: bool = False) -> int:
         """Set charge power."""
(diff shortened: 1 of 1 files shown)


FireSon:
@ahocevar Thanks for the report. I understand the issue, but I do think this is not the correct place to fix it. That is more based on my intentions for these methods, than a complete understanding of the offgrid working of the ZenSDK devices. I know @harrymayr also uses offgrid a lot, and it would have my preference to have a full understanding of the impact of it. Especially if the offgrid is impacting other variables, which lead to unwanted behavior. Based on the complete picture it is easier to find the correct place to fix it.
Do you think that is possible?
Thanks!

custom_components/zendure_ha/device.py
```diff
@@ -735,6 +735,20 @@ async def power_get(self) -> bool:
...
         return await super().power_get()
 
+    async def power_charge(self, power: int) -> int:
+        """Set charge power for ZenSDK devices.
+
+        homeInput is reported as a positive value (grid consumption) for these
+        devices, so the inherited tolerance check (power - homeInput) would
+        never reach zero during charging.  Use power + homeInput instead,
+        which converges to zero when the device has settled at inputLimit=-power.
+        """
```
harrymayr, line 745: This is in general the case, so it must be changed for all devices.
This is a bug (not critical, just set a new power in any case) 
The correct line is:
`        if abs(power + self.homeInput.asInt - self.homeOutput.asInt) <= SmartMode.POWER_TOLERANCE:`
`self.homeOutput.asInt` typically is 0 on calling power_charge (maybe not he first time after a change from discharge)
And it has nothing to do with offGrid power. homeInput == girdInputPower is always a positive value. But power is here negative.

harrymayr: I created a PR #1359 for this bug

ahocevar:
Superseded by #1359, which fixes the same bug at the base-class level.

PR #1348, opened 2026-05-07
Calculate offgrid power instead of trusting an inaccurate sensor

ahocevar:
I'm not sure if it's only my SolarFlow 2400 AC, but the AC offgrid power sensor is very inaccurate (~ 25 W off). So I went ahead and changed the `pwr_offgrid` function to return the calculated offgrid power rather than the sensor reading.

If other users suffer from the same problem, this might be a worthwhile change.

### DIFF
--- a/custom_components/zendure_ha/devices/solarflow2400.py
+++ b/custom_components/zendure_ha/devices/solarflow2400.py
@@ -38,7 +38,7 @@ class SolarFlow2400AC(ZendureZenSdk):
     @property
     def pwr_offgrid(self) -> int:
         """Get the offgrid power."""
-        return self.offGrid.asInt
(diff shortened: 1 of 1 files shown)


FireSon:
@ahocevar Looking at #1348 and #1344 I think they are connected. Can you check that and after we a clear about the best solution for offgrid, merge them into one ?

ahocevar:
Good point, @FireSon. I'll do some testing to find out.

ahocevar:
First observation:

https://github.com/Zendure/Zendure-HA/blob/2c37e46a29474fb32288e3f93b567014cde5cb8a/custom_components/zendure_ha/manager.py#L548-L550

It looks like there is a related, device-specific workaround in `manager.py` already, which may have been an attempt to address the same symptom.

PR #1352, opened 2026-05-08
Add power_discharge deadband

### LINKED ISSUE #1302: [Bug] 1.3.0 upgrade: 2400AC not charging any more, many errors logged
### Home Assistant Version

2026.4.4

v1.3.0

### Describe the bug

After the 1.3.0 update, my SolarFlow 2400 AC does not charge any more, even if my two Hyper 2000 produce enough power. Maybe it's related to these two errors that are logged multiple times every minute:

### To Reproduce

When updating from a working pre-1.3.0 system to 1.3.0, the above error messages start appearing, and 2400AC does not get charged with the excess power from the two Hyper 2000 any more.

See log extracts below

### Expected Behaviour

* 2400AC gets charged with the excess power from the two Hyper 2000.
* The new errors which get logged every couple of seconds disappear atain.

SF2400 AC

### Log Extracts

```shell
Logger: custom_components.zendure_ha.manager
Source: custom_components/zendure_ha/manager.py:414
integration: Zendure Home Assistant Integration ([documentation](https://github.com/zendure/zendure-ha), [issues](https://github.com/zendure/zendure-ha/issues))
First occurred: 12:41:53 PM (33 occurrences)
Last logged: 12:47:07 PM

'ZendureNumber' object has no attribute 'asInt'



Logger: custom_components.zendure_ha.manager
Source: custom_components/zendure_ha/manager.py:415
integration: Zendure Home Assistant Integration ([documentation](https://github.com/zendure/zendure-ha), [issues](https://github.com/zendure/zendure-ha/issues))
First occurred: 12:41:53 PM (33 occurrences)
Last logged: 12:47:07 PM

Traceback (most recent call last): File "/config/custom_components/zendure_ha/manager.py", line 412, in _p1_changed await self.powerChanged(p1, isFast, time) File "/config/custom_components/zendure_ha/manager.py", line 475, in powerChanged await self.power_charge(setpoint, time) File "/config/custom_components/zendure_ha/manager.py", line 565, in power_charge await d.power_charge(-start_pwr - max(0, d.pwr_offgrid) if d.state != DeviceState.SOCFULL else -max(0, d.pwr_offgrid)) File "/config/custom_components/zendure_ha/device.py", line 620, in power_charge return await self.charge(power) ^^^^^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/zendure_ha/device.py", line 741, in charge if power == -SmartMode.POWER_START and self.limitInput.asInt == -SmartMode.POWER_START and self.homeInput.asInt == 0: ^^^^^^^^^^^^^^^^^^^^^ AttributeError: 'ZendureNumber' object has no attribute 'asInt'
```

### Other Information

_No response_

ahocevar:
This pull request replaces #1318.
Fixes #1302.

By adding a `power_discharge` deadband, the AC charging/discharging oscillation in the case of solar surplus stabilizes.

### DIFF
--- a/custom_components/zendure_ha/const.py
+++ b/custom_components/zendure_ha/const.py
@@ -71,3 +71,5 @@ class SmartMode:
 
     POWER_START = 60  # Minimum Power (W) for starting a device
     POWER_TOLERANCE = 5  # Device-level power tolerance (W) before updating
+    CHARGE_DEADBAND_HOLD = 30  # Seconds a small positive setpoint must persist before ramping charge down
+    CHARGE_DEADBAND_GAP = 5  # Seconds of absence from the charge deadband that count as a fresh entry
--- a/custom_components/zendure_ha/const.py
+++ b/custom_components/zendure_ha/const.py
@@ -72,4 +72,3 @@ class SmartMode:
     POWER_START = 60  # Minimum Power (W) for starting a device
     POWER_TOLERANCE = 5  # Device-level power tolerance (W) before updating
     CHARGE_DEADBAND_HOLD = 30  # Seconds a small positive setpoint must persist before ramping charge down
-    CHARGE_DEADBAND_GAP = 5  # Seconds of absence from the charge deadband that count as a fresh entry
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -125,6 +125,8 @@ class ZendureDevice(EntityDevice):
         self.pwr_produced: int = 0
         self.actualKwh: float = 0.0
         self.state: DeviceState = DeviceState.OFFLINE
+        self.charge_deadband_since: datetime | None = None
+        self.charge_deadband_last: datetime = datetime.min
 
         self.create_entities()
 
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -126,7 +126,6 @@ class ZendureDevice(EntityDevice):
         self.actualKwh: float = 0.0
         self.state: DeviceState = DeviceState.OFFLINE
         self.charge_deadband_since: datetime | None = None
-        self.charge_deadband_last: datetime = datetime.min
 
         self.create_entities()
 
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -473,27 +473,27 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                 if setpoint < 0:
                     await self.power_charge(setpoint, time)
                 else:
-                    await self.power_discharge(setpoint)
+                    await self.power_discharge(setpoint, p1)
 
             case ManagerMode.MATCHING_DISCHARGE:
                 # Only discharge, do nothing if setpoint is negative
-                await self.power_discharge(max(0, setpoint))
+                await self.power_discharge(max(0, setpoint), p1)
 
             case ManagerMode.MATCHING_CHARGE | ManagerMode.STORE_SOLAR:
                 # Allow discharge of produced power in MATCHING_CHARGE-Mode, otherwise only charge
                 # d.pwr_produced is negative, but self.produced is positive
                 if setpoint > 0 and self.produced > SmartMode.POWER_START and self.operation == ManagerMode.MATCHING_CHARGE:
-                    await self.power_discharge(min(self.produced, setpoint))
+                    await self.power_discharge(min(self.produced, setpoint), p1)
                 # send device into idle-mode
                 elif setpoint > 0:
-                    await self.power_discharge(0)
+                    await self.power_discharge(0, p1)
                 else:
                     await self.power_charge(min(0, setpoint), time)
 
             case ManagerMode.MANUAL:
                 # Manual power into or from home
                 if (setpoint := int(self.manualpower.asNumber)) > 0:
-                    await self.power_discharge(setpoint)
+                    await self.power_discharge(setpoint, p1)
                 else:
                     await self.power_charge(setpoint, time)
 
@@ -566,7 +566,7 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                     break
             self.pwr_low: int = 0
 
-    async def power_discharge(self, setpoint: int) -> None:
+    async def power_discharge(self, setpoint: int, p1: int) -> None:
         """Discharge devices."""
         _LOGGER.info("Discharge => setpoint %sW", setpoint)
         self.operationstate.update_value(ManagerState.DISCHARGE.value if setpoint > 0 and self.discharge else ManagerState.IDLE.value)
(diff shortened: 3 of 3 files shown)


custom_components/zendure_ha/devices/solarflow2400.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, hass: HomeAssistant, deviceId: str, prodName: str, definition
     @property
     def pwr_offgrid(self) -> int:
         """Get the offgrid power."""
-        return self.offGrid.asInt
+        return self.batteryOutput.asInt + self.homeInput.asInt - self.batteryInput.asInt - self.homeOutput.asInt
```
harrymayr: That is a good idea. I also see that the offGrid power does not fit to gridInputPower (direct grid supply, so battery is not involved). However, only in the logs of the ACE 1500 I see some values for the self consumption of the devices, all others show more or less nothing  🧐

**_Edit: But this is only true for devices, without PV inputs. So it should not used for all devices to calculate offGrid power._**

harrymayr: Maybe not a good idea. I testet it and had during charging some higher batteryInput power than homeInput power which lead to a negative offGridPower, even the offGrid socket was switched of. So there are 2 possibilities:

1. Keep it as it is with self.offGrid.asInt and handle some "unlogic" difference where it is necessary
2. using  `return self.batteryOutput.asInt + self.homeInput.asInt - self.batteryInput.asInt - self.homeOutput.asInt if self.gridOffMode.asInt > 0 else 0`

I would prefer 1.

ahocevar: @harrymayr You mean something like this?
```python
    def pwr_offgrid(self) -> int:
        """Get the offgrid power."""
        offgrid = self.offGrid.asInt
        formula = self.batteryOutput.asInt + self.homeInput.asInt - self.batteryInput.asInt - self.homeOutput.asInt
        # Detect illogic: sensor is zero while formula says otherwise, or they have opposite signs
        if (offgrid == 0 and abs(formula) > 10) or (offgrid > 0 and formula < 0) or (offgrid < 0 and formula > 0):
            unlogic = formula - offgrid
            return offgrid + unlogic
        return offgrid

harrymayr: `offgrid + unlogic = offgrid + formula - offgrid = formula` 😁

harrymayr: But NO. if offGrid == 0 it must return 0. I think this is the truth.

harrymayr: And I think i must correct my formula to
`return self.batteryOutput.asInt + self.homeInput.asInt - self.batteryInput.asInt - self.homeOutput.asInt if self.gridOffMode.asInt != 0 else 0`

ahocevar: I think I'm just going to change pwr_offgrid back to returning `self.offGrid.asInt`. The rest of this pull request is more important, because it fixes a showstopper regression. @harrymayr, are you ok with changing `await d.power_discharge(0 if max(0, d.pwr_offgrid) == 0 else 10)` back to `await d.power_discharge(0)`?

harrymayr: `d.power_discharge(0)` on a device with offGridPower > 0 will change to pass=3 which means `direct grid supply`. Not sure, if this single change will help.
With respect to offGridPower, I have made several changes, which works now pretty well for me (charging my Plug-In car with the help of a SF 2400 AC). 
I'll try to make a PR for that this week, but don't know if @FireSon likes it 😉, it includes several changes.

ahocevar: I removed both the SF2400 offgrid change and the power_discharge change.

Only left the addition of the deadband in the pull request.

ahocevar:
Pushed a follow-up commit (7f66178) addressing a gap I hit while testing this against a multi-device setup.

When a SOCFULL bypass is feeding the home and another device is charging to absorb the surplus, the bypass clamp in `powerChanged` can pin `setpoint` to 0 while `p1` still shows real grid import (the charge is overshooting the surplus). With the current formula `reduction = min(setpoint, battery_charge)` the ramp-down does nothing — `setpoint == 0` ⇒ `reduction == 0` ⇒ `power_charge(-(battery_charge + offgrid_floor))` re-issues the existing inputLimit and the import stays.

The follow-up drives the reduction from observed `p1` instead: at least `POWER_START` per cycle while import persists, scaling up with the actual import so large overshoots converge quickly, and holding steady at `p1 <= 0` so we don't oscillate when balanced. `power_discharge` now takes `p1` as a parameter; all call sites in `powerChanged` updated.

ahocevar:
Pushed 8a3366a after observing the deadband still didn't fire in practice.

Root cause: `CHARGE_DEADBAND_GAP = 5s` resets `charge_deadband_since` on every iteration whose interval exceeded the gap. Actual update cadence in my logs is 4–15s (TIMEZERO is 4s, but p1-sensor latency stretches intervals), so the 30s hold never accumulated and `power_charge` was never re-issued.

Switched to a None/Some sentinel for `charge_deadband_since` — set on entry, cleared on exit, also cleared for any device that drops out of `self.charge` between iterations. `CHARGE_DEADBAND_GAP` and `charge_deadband_last` are removed since the time-gap heuristic is no longer used.

ahocevar:
Squashed the previous `max(p1, POWER_START)` floor into the p1-feedback commit (it was overshooting into export and re-triggering charge mode every 30s — the same oscillation #1352 was meant to fix). Reduction is now `min(p1, battery_charge)` so the ramp lands on balance. Branch force-pushed; history is now two follow-up commits on top of the original.

ahocevar:
Pushed edf5038 — the oscillation came back from an interaction with the hysteria/`charge_time` logic, not the deadband itself.

Trace from a multi-flip cycle in my logs:

1. Charge mode ramps the absorber up to ~700W.
2. p1 briefly turns positive; bypass clamp pins setpoint to 0; `power_discharge` runs.
3. `power_discharge` resets `charge_time = datetime.max` unconditionally.
4. Deadband fires after 30s and trims charge by p1 (working as intended).
5. p1 swings negative; `power_charge` runs, sees `charge_time == datetime.max`, treats it as a fresh activation, sets a 60s hysteria, and forces setpoint=0 → `Power charge … => 0`. Active charge dropped entirely.
6. Surplus piles up, 60s later the device ramps back to ~800W. Tiny positive p1 → discharge → reset → repeat.

So every zero-crossing was retriggering the 60s blackout. Fix: only reset `charge_time` when `setpoint >= POWER_START`, so the bypass-clamped deadband region doesn't count as exiting charge mode.

FireSon:
Thanks for the careful work here, and for the #1318 background. The principle is sound: don't flip a charging device into discharge mode just because the controller's setpoint briefly crosses zero. Hold for a moment, ramp down using observed `p1`, stay in charge mode at reduced `inputLimit`. That's the right instinct for #1302.

I'm going to defer this for the current release. Two reasons:

- The change carries a fair amount of new surface (per-device state, hardcoded 30s hold, signature change across six call sites, hysteria-reset change, acMode-pattern change). I want to keep the current release small.
- The next version will use a different power distribution approach (alongside HEMS-commands and real-local-MQTT). The deadband logic would interact with that new distribution, and I'd rather design it together once the new distribution is settled.

Your diagnosis and reasoning here are exactly the input I'll want at that point. I'll open a tracking issue so this doesn't get lost.

PR #1359, opened 2026-05-10
Fix power charge logic

harrymayr:
In power_charge, power is a negative value, but homeInput and homeOutput are always positive


ahocevar:
I can confirm that this fixes a concrete symptom on my system (SF 2400 AC overcurrent on the off-grid plug, manager re-issuing charge() on every tick because the tolerance check never converges).

PR #1363, opened 2026-05-11
Random start charge to prevent no start issue (SF2400AC+)

zoic21:
Add random number (between 0 to 10) on start charge to prevent no charging issue on SF2400AC+

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -7,6 +7,7 @@ import hashlib
 import json
 import logging
 import traceback
+import random
 from collections import deque
 from collections.abc import Callable
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


FireSon:
Thanks for the PR @zoic21.

We're going to close this in favour of a different approach. A similar random-jitter fix existed before in commit `f787a6a` ("Add randomness to start power") but was deliberately removed in `467e7a4` in favour of a deterministic anti-oscillation guard in `ZendureZenSdk.charge` / `discharge` (`custom_components/zendure_ha/device.py`), which nudged the limit by ±10 W when the device failed to latch on the first request.

Master now extends that guard into a cumulative walk (commit `58ac014`): each 4 s tick where the manager asks for a start and the device still shows no `homeInput` / `homeOutput`, the limit walks by 4 W in the start direction, capped at ±120 W, and self-resets the moment the device starts drawing. Worst-case ramp 60 to 120 W is about 60 s.

This addresses the same SF2400AC+ no-start symptom but deterministically. It is observable in logs, has no overshoot once the device latches, and adds no per-call randomness in the manager.

If you still see the issue on master after this, please open an issue with logs showing `limitInput` / `homeInput` over time during a failed start and we'll dig further.

Issue #1322, opened 2026-04-29
[Bug] Only one device in bypass, the other in Standby


JD122:
### Home Assistant Version

2026.4.4

### Zendure Integration Version

v1.3.1

### Describe the bug

Together with the last update I got another zendure device. So I can't say if it's an problem with the last update. I have an Solar Flow 800 pro and 800 Plus. 
As soon as the Battery was fully charged, happened on both devices nearly at the same Time, one device goes into bypass and the other one goes into standby an does nothing. 
The one in bypass seems to only supply half the power needed for the home. 
It also can change which one of the devices goes into standby. 

I would expect that if the house needed 300w both would supply 150w. 
I will attach pictures from the Home wattage consumption together with what the zendure outputs. The white line is the house Consumption. It can clearly bee wenn that as soon as both batteries are full, only half of the needed power is supplied.

I use the Smart mode und my smart meter is the Shelly 3m.

<img width="1080" height="1002" alt="Image" src="https://github.com/user-attachments/assets/6627bd34-ea25-4d1c-9aaf-232d5f697baa" />

<img width="1080" height="1931" alt="Image" src="https://github.com/user-attachments/assets/7b4507a0-acae-415c-91c7-6bcfcd7df2bf" />

### To Reproduce

use two zendure Solar Flows
set smart mode for the integration 
wait until both devices are charged 
only one device supplies power to the house and only have the power needed 

### Expected Behaviour

both devices supply power and all power needed 

### What device are you using?

SF800 Pro

### Log Extracts

```shell

```

Juju-no-dev:
Same issue with SF800 Plus + SF2400 Pro combination
I have a SF800 Plus and a SF2400 Pro in the same fuse group (SF800 Plus configured as "Part of SolarFlow 2400 Pro fusegroup", max 3600W).
After updating from 1.2.7-pre4 to 1.3.0/1.3.1, I noticed the SF800 Plus discharging alone (down to 25%) while the SF2400 Pro stays in standby at 94% SOC. Before the update, both devices discharged in parallel as expected.
Environment:

Home Assistant OS 17.2
Home Assistant Core 2026.4.4
Zendure HA Integration 1.3.1

Setup:

SF800 Plus (1x AB2000 battery)
SF2400 Pro (2x AB3000L batteries, 5400 Wh)
Zendure Manager in smart mode
ZenSDK local connection (no HEMS)
Grid export: allowed

[...]

PR #1368, opened 2026-05-11
Classify devices with active off-grid load as discharge candidates

ahocevar:
## Problem

On single-port SolarFlow devices — confirmed on the SF 2400 AC, likely the same family-wide — `outputHomePower` reads 0 and `gridInputPower` also reads 0 even while the device is autonomously back-feeding the grid (verified with an external power sensor on the on-grid plug). The current classifier in `ZendureManager.powerChanged`:

```python
elif (home := d.homeOutput.asInt) > 0:
    self.discharge.append(d)
```

sends such a device to `idle`. With `setpoint < POWER_START`, the discharge loop's idle-wake-up never fires (`dev_start = 0`), so the device sits unaddressed while small residual home demand is supplied by the grid.

In a real session this stuck state persisted for 17+ minutes with the device exporting ~13 W externally but reporting `homeOutput=0`, `homeInput=0`, `pwr_offgrid=22 W`, `acMode=1`.

## Fix

Treat `pwr_offgrid > 0` as a reliable "device is running" signal and route the device into `discharge` even when `homeOutput` reads 0:

```python
elif (home := d.homeOutput.asInt) > 0 or (d.pwr_offgrid > 0 and d.state != DeviceState.SOCEMPTY):
```

`home` stays 0 in the new branch, so `setpoint` is not bumped by an unreliable sensor — we don't pretend to know an export figure that the firmware isn't giving us. The discharge distribution then issues `acMode=2` and the battery covers off-grid + home demand.

## Scope

- Only affects the SolarFlow family that overrides `pwr_offgrid` (SF 800, SF 1600, SF 2400 variants). All other devices (Hyper, AIO, ACE, hubs, SuperBase) inherit `pwr_offgrid = 0` from the base class and are unaffected.
- For SolarFlow devices where `homeOutput > 0` works correctly, the first branch fires first and accounting is unchanged.
- The new branch only fires when `homeOutput == 0` but `pwr_offgrid > 0` — i.e., the gap case where the firmware under-reports.
- `state != SOCEMPTY` guard mirrors the rationale of the surrounding block: don't try to discharge a flat battery.

## Notes

- Independent of #1352 (charge-side deadband) — they compose cleanly: #1352 smooths the charge ramp-down, this PR handles the post-rampdown state when off-grid load remains active.
- Independent of #1359, which touches sign arithmetic in `power_charge`'s no-action check.
- This is a workaround for a firmware reporting bug. The "right" fix would be a sensor we can trust for on-grid back-feed direction; absent that, `pwr_offgrid` is the next-best generic signal.

## Test plan

- [x] Reproduced the stuck-idle state on SF 2400 AC; with the patch the device gets classified as `discharge` and is commanded `acMode=2`, battery covers home demand and off-grid load
- [ ] Confirm no regression on Hyper-only setups (unaffected by construction since `pwr_offgrid == 0`)
- [x] Confirm no regression on SF 800 / SF 1600 setups where `homeOutput` was already populating correctly

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -447,6 +447,14 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                     self.discharge_produced -= d.pwr_produced
                     self.discharge_weight += d.pwr_max * d.electricLevel.asInt
                     setpoint += home
+                    # When a device is in discharge only because of pwr_offgrid (no home output),
+                    # it can still absorb surplus from another device's SOCFULL bypass. Expose it
+                    # as an idle candidate too, so power_charge can engage it before a sibling
+                    # device starts charging autonomously from AC.
+                    if home == 0 and d.state != DeviceState.SOCFULL:
+                        self.idle.append(d)
+                        self.idle_lvlmax = max(self.idle_lvlmax, d.electricLevel.asInt)
+                        self.idle_lvlmin = min(self.idle_lvlmin, d.electricLevel.asInt)
 
                 else:
                     self.idle.append(d)
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -436,8 +436,10 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                     self.charge_optimal += d.charge_optimal
                     self.charge_weight += d.pwr_max * (100 - d.electricLevel.asInt)
                     setpoint += -d.homeInput.asInt  # use gridInputPower directly; offgrid consumers are invisible to P1
-                # SOCEMPTY means, it could not discharge the battery, but it is still possible to feed into the home using solarpower or offGrid
-                elif (home := d.homeOutput.asInt) > 0:
+                # SOCEMPTY means, it could not discharge the battery, but it is still possible to feed into the home using solarpower or offGrid.
+                # Also classify as discharge when the device has an off-grid load: on devices like the SF 2400 AC, outputHomePower reads 0
+                # even while the device is autonomously back-feeding the grid, so pwr_offgrid is our only reliable "device is running" signal.
+                elif (home := d.homeOutput.asInt) > 0 or (d.pwr_offgrid > 0 and d.state != DeviceState.SOCEMPTY):
                     self.discharge.append(d)
                     self.discharge_bypass -= d.pwr_produced if d.state == DeviceState.SOCFULL else 0
                     self.discharge_limit += d.fuseGrp.discharge_limit(d)


Juju-no-dev:
Confirming this exact bug on SF2400 Pro (running alongside an SF800 Plus). The issue occurs regardless of fuse-group configuration. Observed both when devices share the same fuse group and when split across separate phases.
Important context on regression: I never encountered this bug on 1.2.7-pre4. It only started after upgrading to 1.3.0. I then rolled back to 1.2.7-pre4 (per #1322), but the freeze now persists even on this older version (no firmware update on the Zendure side in between). I encountered the bug several times but it took me a while to figure out the actual cause of the freeze.
Same signature as described:

output_home_power = 0 W, solar_input_power = 0 W, pack_input_power = 0 W (all frozen)
output_limit collapsed to 50 W
battery_state = Standby, bypass = OFF
Lasted 1h12 on 2026-05-09 (17:12 → 18:24) before manual recovery
During the freeze, the SF800 Plus keeps injecting while the sun is up, then discharges alone at nightfall to cover house demand

It happened again today (2026-05-11 at 18:32 Paris time), but this time my HA automation caught it: detection within 5 min on the signature bypass=off + output_limit≤100W + SOC≥99%, then recovery sequence (manager=off → ac_mode=input → wait 20s → ac_mode=output → manager=smart). Device recovered within ~40 seconds, solar back to 232W. This confirms the device itself is healthy, the integration just stops commanding it once homeOutput reports 0.
This PR addresses the root cause exactly. Will test as soon as it lands in a release.
Setup: HA 2026.4.4, Zendure HA 1.2.7-pre4, SF2400 Pro + SF800 Plus in Part of SolarFlow 2400 Pro fusegroup mode, ZenSDK local MQTT, grid_reverse=allow.

ahocevar:
Pushed a follow-up commit (f910354) that closes a regression this PR introduces in multi-device setups.

The classifier now routes any device with `pwr_offgrid > 0` into `discharge`. If such a device has `homeOutput == 0` it is in discharge only because of off-grid load — its battery still has spare capacity. But because it is no longer in `idle`, `power_charge`'s idle-wake-up cannot select it as an absorber, so any surplus produced by a sibling SOCFULL device's bypass goes elsewhere (in practice: a non-full sibling decides to charge autonomously from AC).

The addendum keeps the discharge classification intact and additionally exposes such devices in `self.idle` (only when `home == 0` and `state != SOCFULL`), restoring their charge-candidate eligibility. The SOCEMPTY/offgrid case this PR fixes is unaffected.

Juju-no-dev:
This matches a behavior I observed across multiple occurrences with my setup. When the SF800 Plus battery reached 100% SOC, its surplus solar (its own panels still producing in bypass) correctly fed into the SF2400 Pro's battery. But the reverse never worked: when the SF2400 Pro battery reached 100% first, its surplus solar didn't charge the SF800 Plus. The SF800 Plus kept charging only from its own panels, while the SF2400's solar surplus went to the grid instead. This matches exactly the asymmetry your commit f910354 describes and fixes. Looking forward to the release.

FireSon:
Thanks for the detailed scenario walk-through, the scope notes, and especially for documenting both the bug fix and the optimization in one place. Both pieces target real problems.

I'd like to merge the OR-condition first as a clean stand-alone fix:

```python
elif (home := d.homeOutput.asInt) > 0 or (d.pwr_offgrid > 0 and d.state != DeviceState.SOCEMPTY):
```

Verified narrow: only the SF 800/1600/2400 family override `pwr_offgrid`, the base class returns 0, so other devices are unaffected by construction. `home` stays 0 in the new branch so `setpoint` is not bumped by an untrustworthy figure. That fixes the 17-minute stuck-idle case directly.

On the dual-listing block: the absorption case you're targeting is a real win and I'd like to solve it, but the implementation has a problem I'd like you to address. Listing the device in both `self.discharge` and `self.idle` means `power_charge` touches it twice whenever setpoint is below -POWER_START (which is exactly the surplus-to-absorb case): first `d.power_discharge(-10)` at line 513 because it's in `self.discharge`, then `d.power_charge(-60 - pwr_offgrid)` at line 564 because it's in `self.idle`. End state is correct, but `acMode` flaps from 2 to 1 between the commands and `outputLimit` is briefly -10, which means double-traffic per cycle and a contradictory state if the second command is delayed or dropped. It also breaks the assumption elsewhere in the dispatcher that `discharge` and `idle` are disjoint.

Would you take a look at how to engage the device for absorption without putting it in both lists? Happy to merge the OR-condition now and review the absorption case as a follow-up PR with whatever shape you land on.

ahocevar:
@FireSon Thanks for taking a look at this! I think just merging the OR condition for now is a clean split. I'll try to come up with a follow-up pull request in the next couple of days.

PR #1403, opened 2026-05-25
HUB/AIO: Add device minimum discharge power

igosoft:
## Description

This PR introduces a configurable **Min Discharge** slider for each device to prevent inverter MPPT cooldowns and ensure instant responsiveness. 

### Why this is needed
Previously, the inverter could build up to ~15 minutes delay by:
* the microinverter's grid-reconnect (~5 min)
* auto-restart (~10 min) timers 
after it goes idle (documented in the Hub 2000 manual).
By allowing users to enforce a minimum baseline output power, the inverter stays continuously active and ready to perform.

### Key Changes
* **Min Discharge Slider**: Added a slider ranging from `0–800W`.
* **Finer Low-Power Control**: Configured dynamic steps of `30W` when the value is below `100W`.
* **Floor Enforcement**: The manager now strictly enforces this minimum limit during both active discharge and idle device startup phases.
* **Mode Exclusions**: The minimum output floor is automatically bypassed in the following modes:
  * `STORE_SOLAR`: Prioritizes full power transfer directly to the batteries.
  * `MANUAL`: Defers entirely to user-defined manual output control.

---

## Behavior Example

| Current Power P1 | Min Output Setting | Expected Final Output |
| :--- | :--- | :--- |
| 0W | 100W | **100W** |
| 200W | 100W | **200W** |
| -300W | 100W | **-300W** |

---

## How to Test
1. Open the device settings (eg HUB2000) and verify the new **Min Discharge** slider works (0-800W).
2. Verify that values below 100W change in steps of 30W.
3. Set the slider to 100W, trigger an idle startup, and ensure the power output does not drop below 100W.
4. Switch to `STORE_SOLAR` or `MANUAL` mode and confirm the minimum limit is ignored.

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -140,6 +140,7 @@ class ZendureDevice(EntityDevice):
         self.actualKwh: float = 0.0
         self.state: DeviceState = DeviceState.OFFLINE
         self.exports_bypass: bool = True
+        self.awake: bool = False
 
         self.create_entities()
 
@@ -286,6 +287,22 @@ class ZendureDevice(EntityDevice):
 
         soc = self.minSoc.asNumber
         return 0 if level <= soc else min(999, self.kWh * 10 / power * (level - soc))
+
+    @property
+    def min_output(self) -> int:
+        # Minimum discharge power to keep the microinverter engaged; only HUB/AIO expose it.
+        entity = getattr(self, "minOutputPower", None)
+        return entity.asInt if entity is not None else 0
+
+    def on_direction_change(self, charging: bool) -> None:
+        # Awake while discharging so power_discharge can hold the microinverter above its minimum.
+        self.awake = not charging
+
+    def localEntityWrite(self, entity: EntityZendure, value: Any) -> None:
+        # Snap values below 101 to the nearest step in {0, 30, 60, 90, 100}.
+        if entity.propertyName == "min_output_power" and isinstance(value, (int, float)) and value <= 100:
+            value = min([0, 30, 60, 90, 100], key=lambda x: abs(x - value))
+        entity.update_value(value)
 
     async def entityWrite(self, entity: EntityZendure, value: Any) -> None:
         if entity.translation_key is None:
@@ -655,6 +672,8 @@ class ZendureDevice(EntityDevice):
     async def power_discharge(self, power: int) -> int:
         """Set discharge power."""
         power = max(0, min(power, self.discharge_limit))
+        if self.awake and self.min_output > 0 and self.state not in (DeviceState.SOCEMPTY, DeviceState.SOCFULL) and self.electricLevel.asInt > self.minSoc.asNumber:
+            power = max(power, self.min_output)
         if abs(power - self.homeOutput.asInt + self.homeInput.asInt) <= SmartMode.POWER_TOLERANCE:
             _LOGGER.info("Power discharge %s => no action [power %s]", self.name, power)
             return self.homeOutput.asInt
--- a/custom_components/zendure_ha/devices/aio2400.py
+++ b/custom_components/zendure_ha/devices/aio2400.py
@@ -3,10 +3,12 @@
 import logging
 from typing import Any
 
+from homeassistant.components.number import NumberMode
 from homeassistant.core import HomeAssistant
 
 from custom_components.zendure_ha.const import SmartMode
 from custom_components.zendure_ha.device import ZendureLegacy
+from custom_components.zendure_ha.number import ZendureRestoreNumber
 
 _LOGGER = logging.getLogger(__name__)
 
@@ -18,6 +20,7 @@ class AIO2400(ZendureLegacy):
         """AIO 2400 cannot charge using AC"""
         self.setLimits(0, 1200)
         self.maxSolar = -1200
+        self.minOutputPower = ZendureRestoreNumber(self, "min_output_power", self.localEntityWrite, None, "W", "power", 800, 0, NumberMode.SLIDER)
 
     async def charge(self, power: int) -> int:
         _LOGGER.info("No AC charge for %s available", self.name)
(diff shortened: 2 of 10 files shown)


harrymayr:
> the inverter could take up to 30 minutes to warm up its MPPT algorithm after being idle

Can you explain, what's the issue with the warm-up? What does not work if the device is not warmed up?
I wonder if this is more (only?) a problem of the HUB-family and maybe of the SF 800. Not sure, if we should add this feature to all devices.

igosoft:
The issue originates from the inverter behavior. When there is no DC input for a period of time, the inverter's MPPT takes up to 30 minutes to 'warm up' and restart. This delay causes the Zendure HUB 2000 (family) output to be blocked by the inverter. As a result, no power is delivered to the home grid, even when the household demand (P1) is high. In the context of the issue, yes it's just HUB family problem. 

This feature (now) isn't just a workaround for my specific case. While trying to fix this issue, I realized it shouldn't be resolved directly within the HUB 2000's context.  I propose adding a 'minimum output' property for battery discharge modes. This feature would benefit all users including owners of HUB2k HUB1.2k too by keeping the inverter's MPPT continuously active, completely eliminating these startup delay issues. 

Do you think this feature should be limited to the HUB family?

```

+-----------------------+          +-----------------------+

|    Solar Panels PV    |          |    Solar Panels PV    |
+-----------+-----------+          +-----------+-----------+

            |                                  |
         [DC] Solar Power                   [DC] Solar Power
            v                                  v
+----------------------------------------------------------+

|                    ZENDURE HUB 2000                      |
|               (Smart Energy Management)                  |
+---------------------+---------------------+--------------+

                      |                     |
             [Battery]  Charge / Discharge
                      v                     v
            +---------------------------------+

            |          Zendure Battery        |
            +---------------------------------+

                      |
                      | (Managed DC Output)
                      v
            +---------------------------------+

            |     DC Y-Splitter Cables        |
            +--------+----------------+-------+

                     |                |
         [DC] Input 1|                | [DC] Input 2
                     v                v
+--------------------+----------------+--------------------+

|               HOYMILES HMS-2000 (Inverter)               |
|  +------------------------+  +------------------------+  |
|  |       MPPT Input 1     |  |       MPPT Input 2     |  |
|  +------------------------+  +------------------------+  |
+---------------------------+------------------------------+

                            |
                         [AC] Grid Power (230V)
                            v
                    +---------------+
                    | Home Grid Wall|
                    |    Outlet     |
                    +---------------+


```

CounterChip:
As far as i know this problem makes fully sense for the hub series. 
I saw this behavior with some inverters. 
For example with a HMS-2000 like igosoft mentioned.
When there is some voltage but not enough power, the inverter shuts down. 
In the cases where i was able to see it, it did a retry of the MPPT-start again after 15 minutes and it doesnt care if you get above the minimum voltage of the mppt again shortly after a "failed" start.

I think you understand how it behaves with some graphs.
This example is from a HMS-2000.
Voltage from panels:
grafik
Power from inverter (purple) and DC power panels Panels same colours as above:
grafik
As you can see it tries a start, because there is enough voltage. It hadnt enough power at 08:05, where it failed to start and then started again at 08:20, where it had enough power available.

To stay active probably around 10W-15W could be sufficient for these inverters.
Hope this helps understanding it a little bit easier.

harrymayr:
@CounterChip, @igosoft  Thanks for the explanation.

> Do you think this feature should be limited to the HUB family?

Yes, because only these devices are conneted to a micro-inverter. All other devices are connected directly to the home grid.

igosoft:
Sounds reasonable. However, (also) as a non-HUB family user, I would also like to use it on other devices. 

I'm going to restrict this property to the HUB family asap. If needed, it can be migrated to the rest of the devices later

PS
You probably meant that other devices have an inverter built-in

FireSon:
Hi @igosoft — I pushed a reworked version of your min-discharge feature onto this branch (force-push) and would appreciate your review before we merge.

**What changed vs your original:**
- **Scoped the entity to HUB/AIO only** — `min_output_power` is now added just in `hub1200.py`, `hub2000.py`, `aio2400.py`. No other device gets the setting (that was the main concern: settings appearing on devices that don't need them).
- **Enforcement moved out of the shared manager loop** into the device's own `power_discharge` override: `if self.awake and self.min_output > 0: power = max(power, self.min_output)`. This keeps the shared distribution code untouched, so non-HUB/AIO devices are provably unaffected.
- **`awake` is driven by a charge/discharge direction signal**: the manager calls `_set_direction()` at the existing charge/discharge transition points (and on OFF), which calls `on_direction_change()` on every device. Base hook is a no-op.
- **Dropped the `SMART_SOLAR_PASSTHROUGH` mode** — that's now handled by the `MATCHING_DISCHARGE` change already merged in #1408, so a separate mode isn't needed.

**Rationale note:** I'd suggest rewording the "30-minute MPPT warm-up" justification — MPPT itself converges in seconds. The real delay is the microinverter's grid-reconnect (~5 min) + auto-restart (~10 min) timers after it goes idle (documented in the Hub 2000 manual). Keeping it engaged avoids those. Same feature, accurate reasoning.

**Not addressed (open questions for you):**
- No SOC/reserve guard: while discharging, a HUB/AIO is held at min even at low SoC. Worth a floor-disable below a reserve?
- Idle HUB/AIO not in the discharge list won't be re-commanded every tick — the floor applies while it's actively in discharge.

Please review the approach and test on a real HUB/AIO. Thanks for the feature and the analysis.

igosoft:
@FireSon 

> No SOC/reserve guard: while discharging, a HUB/AIO is held at min even at low SoC. Worth a floor-disable below a reserve? 

I added the min SoC guard

> Idle HUB/AIO not in the discharge list won't be re-commanded every tick — the floor applies while it's actively in discharge.

I prefer idle devices to be taken into account if a min output is set (this is extremely useful with two HUB2k units connected to one inverter and power demand peaks). However, I'm not sure if preventing a device from entering the IDLE state has any negative impact on that device

igosoft:
with the latest changes min output feature doesn't work yet as intended (from time to time). I'm trying to find out why

igosoft:
Merged to https://github.com/Zendure/Zendure-HA/pull/1567 due to missing tests and high complexity

PR #1408, opened 2026-06-01
New manager mode (Solar Pass-through) to smart pass-through all power to house 

igosoft:
## Description

Adds a new smart mode: **Solar Pass-through** (`smart_solar_passthrough`), designed for situations where selling excess PV energy to the grid is profitable.

In this mode, all PV production is passed directly to the house and grid — the battery is only used to cover any remaining P1 grid import demand

  | PV | P1 (demand) | Battery output| Grid (export) |
  |-----|-------------|---------|---------------|
  | 300W | 100W | 0W | −200W |
  | 100W | 100W | 0W | 0W |
  | 100W | 200W | 100W | 0W |
  | 500W | 100W | 0W | −400W |

  **Useful when:**
  - Energy export prices are high and uers want to maximize grid export from PV
  - Users want the battery to act only as a P1 safety net, not a grid feed device

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -491,8 +491,8 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                     await self.power_discharge(setpoint)
 
             case ManagerMode.MATCHING_DISCHARGE:
-                # Only discharge, do nothing if setpoint is negative
-                await self.power_discharge(max(0, setpoint))
+                # Discharge to cover demand and always pass through available solar; never charge
+                await self.power_discharge(max(self.produced, setpoint))
 
             case ManagerMode.MATCHING_CHARGE | ManagerMode.STORE_SOLAR:
                 # Allow discharge of produced power in MATCHING_CHARGE-Mode, otherwise only charge


custom_components/zendure_ha/const.py
```diff
@@ -42,6 +42,10 @@ class ManagerMode(Enum):
     MATCHING_DISCHARGE = 3
     MATCHING_CHARGE = 4
     STORE_SOLAR = 5
+    SMART_SOLAR_PASSTHROUGH = 6
+
+
+DISCHARGE_MODES = frozenset({ManagerMode.MATCHING, ManagerMode.MATCHING_DISCHARGE, ManagerMode.MATCHING_CHARGE, ManagerMode.SMART_SOLAR_PASSTHROUGH})
```
igosoft:
```suggestion
DISCHARGE_MODES = frozenset({ManagerMode.MATCHING, ManagerMode.MATCHING_DISCHARGE, ManagerMode.SMART_SOLAR_PASSTHROUGH})
```

FireSon:
Hi @igosoft, thanks for the contribution. Two things I'd like to discuss before merging.

**First, a higher-level concern.** We already have six manager modes (off, manual, smart, smart-discharge, smart-charge, store solar). Adding a seventh raises the cognitive load on users and on us as maintainers, and from the code below I'm not yet convinced this mode does something the existing ones can't. I'd like to be conservative about adding modes unless the gap they fill is clearly distinct and meaningful. So I want to understand the design before saying yes.

**Second, the specifics.** Looking at the code I can't reconcile the behavior with the description table.

*How is this different from `MATCHING_DISCHARGE` + `gridReverse=allow`?* Both modes are discharge-only, both leave the battery as a backup, and in both cases the device autonomously exports PV when battery is full and export is allowed. Walking through your "PV=500W, P1=-400W" row:

- `MATCHING_DISCHARGE`: setpoint after credits ~= 0, so `power_discharge(0)`. Battery idle. PV exports autonomously. Grid = -400W. ✓
- `SMART_SOLAR_PASSTHROUGH`: `power_discharge(max(500, 0)) = power_discharge(500)`. Device target 500W, PV provides 500W, battery contributes 0. Same end state.

Could you describe a scenario where the two modes produce a different observable outcome? If it's about not trusting the device's autonomous export behavior, I'd like to see that documented in the code.

*The "house demand exceeds PV" row of your table doesn't seem to match the formula.* For PV=100W, P1=200W (importing) you list:

| PV | P1 | Battery | Grid |
| 100W | 200W | 100W | 0W |

The formula computes `power_discharge(max(100, max(0, 200))) = power_discharge(200)`. With device target = 200W and PV providing 100W of that, the battery contributes 100W. Total device output to home = 200W. House demand = 100 + 200 = 300W. So home pulls 100W from the grid, giving Grid = +100W, not 0W.

If the intent really is "battery covers all remaining P1 demand", shouldn't the formula be `power_discharge(max(self.produced + max(0, p1), 0))` or similar, so the battery target is added on top of PV rather than max-of?

If we can land on a clear answer to both points (gap vs MATCHING_DISCHARGE, and table-matches-formula), I'm happy to take this in. Otherwise I'd prefer to keep the mode count down.

igosoft:
Hi @FireSon 

Thanks for the comments and questions. I also found that formula problematic**, so** I started working on tests first to confirm it works across all scenarios with multiple devices on a single phase. The reason I needed the passthrough mode was that MATCHING_DISCHARGE didn't perform as expected. I'm going to modify my automation to use MATCHING_DISCHARGE and will follow up with some data

| PV Output (W) | P1 (demand) (W) | Battery output (W) | Grid (export) (W) | MATCHING_DISCHARGE max(0, P1 + home)   | SMART_SOLAR_PASSTHROUGH max(produced, max(0, P1))  |
| :---: | :---: | :--- | :--- | :--- | :--- |
| 200 | 100 | MD: 100 / SSP: 0 | MD: 0 / SSP: 100 | max(0, 100+200) = 300 | max(200, max(0,100)) = 200 |
| 50 | 100 | MD: 100 / SSP: 50 | MD: 0 / SSP: 50 | max(0, 100+50) = 150 | max(50, max(0,100)) = 100 |
| 100 | -100 | MD: 0  / SSP: 0 | MD: 0 / SSP: -100 | max(0, -100+100) = 0 | max(100, max(0,-100)) = 100 |
| 400 | -300 | MD: 0  / SSP: 0 | MD: 0 / SSP: -300 | max(0, -300+400) = 100 | max(400, max(0,-300)) = 400 |

custom_components/zendure_ha/manager.py
```diff
@@ -481,6 +481,10 @@ async def powerChanged(self, p1: int, isFast: bool, time: datetime) -> None:
                 # Only discharge, do nothing if setpoint is negative
                 await self.power_discharge(max(0, setpoint))
 
+            case ManagerMode.SMART_SOLAR_PASSTHROUGH:
+                # Pass all solar to house; use battery only to cover remaining P1 demand
+                await self.power_discharge(max(self.produced, max(0, p1)))
```
igosoft:
```suggestion
                await self.power_discharge(max(self.produced, setpoint))
```

igosoft:
Hi @FireSon 
I have just finished implementing the tests for the Manager class. I can confirm that if we modify MATCHING_DISCHARGE to the following, the SMART_SOLAR_PASSTHROUGH case isn't needed
```
ManagerMode MATCHING_DISCHARGE
await self.power_discharge(max(self.produced, setpoint))
```

I'm going to raise a PR with the fix unless you have some spare time before I do

FireSon:
Thanks @igosoft — took you up on the offer and pushed your proposed fix onto this branch, so we don't need a second PR.

The PR now drops the separate `SMART_SOLAR_PASSTHROUGH` mode entirely and instead changes `MATCHING_DISCHARGE` to `power_discharge(max(self.produced, setpoint))`. That passes through all available solar and lets the battery cover only demand above solar (never draining for export, since power_discharge uses produced power first). It also fixes the import-case behaviour the original separate mode got wrong, and keeps the mode count down.

Merging. Thanks for the analysis and the Manager-class tests that confirmed it.

igosoft:
@FireSon thanks

D4MnEd:
Hi @FireSon & @igosoft,
Thanks a lot for your work but, if I understand the benefit of pass-through, the purpose of the smart-discharge mode was to charge the battery using surplus solar power.
The value of this mode is therefore more limited now (or am I wrong ?), since it’s no longer possible to charge the battery using this mode.

Tomakava:
I also have problem with this mode when Energy Export set to forbidden. I want to control grid power exactly using P1. Now zero export is no longer possible.

PR #1422, opened 2026-06-10
fix(manager): cap discharge_bypass per device to prevent setpoint clamp from cancelling active charging (regression of #1151 fix)

zoic21:
# Bug
When one device is charging surplus while a SOCFULL device bypasses a few watts to the home, the setpoint clamp from the #1151 fix forces the setpoint to 0 and cuts the charge. The hysteresis then blocks recharging for 60 s, so the charge cycles 0 ↔ full power about every minute as long as the full device bypasses.
```
Device power SolarFlow 2400 AC+ => state:CHARGE homeInput:1248W batteryInput:1248W ...
Device power SolarFlow 2400 Pro => state:IDLE homeOutput:22W ... DeviceState.SOCFULL
P1 ======> p1:2 isFast:False, setpoint:0W stored:22W   <-- real setpoint is -1246W
Discharge => setpoint 0W                               <-- charge cut, 60s penalty
```

# Cause
For a SOCFULL device, homeOutput is added to the setpoint but -pwr_produced is subtracted via discharge_bypass. When these diverge (trickle charge, sensor skew), more is subtracted than was added, creating a phantom negative setpoint. The max(0, ...) clamp neutralizes that phantom, but it cannot tell it apart from a real negative setpoint (an ongoing charge) and erases both.

# Fix
Cap the per-device bypass at the homeOutput actually added to the setpoint. The phantom can no longer exist, so the clamp is removed.

```
# before (classification, discharge branch)
self.discharge_bypass -= d.pwr_produced if d.state == DeviceState.SOCFULL and d.exports_bypass else 0

# after
if d.state == DeviceState.SOCFULL and d.exports_bypass:
    self.discharge_bypass += min(-d.pwr_produced, home)
```

```
# before
if self.discharge_bypass > 0:
    setpoint = max(0 if p1 >= 0 else setpoint - self.discharge_bypass, setpoint - self.discharge_bypass)

# after
setpoint -= self.discharge_bypass
```

The #1151 case is unchanged: with prod 150 W / home 100 W / p1 = 0, the capped bypass gives +100 − 100 = 0, same as the clamp. Since the capped bypass never exceeds the added homeOutput, the setpoint can only go negative when a real charge is in progress — exactly when it must.

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -441,7 +441,12 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                 # SOCEMPTY means, it could not discharge the battery, but it is still possible to feed into the home using solarpower or offGrid
                 elif (home := d.homeOutput.asInt) > 0:
                     self.discharge.append(d)
-                    self.discharge_bypass -= d.pwr_produced if d.state == DeviceState.SOCFULL and d.exports_bypass else 0
+                    # Cap the bypass at the homeOutput actually added to the setpoint for this
+                    # device: pwr_produced can exceed homeOutput (internal trickle charge, sensor
+                    # skew), and subtracting more than was added fabricates a phantom negative
+                    # setpoint — the root cause of #1151.
+                    if d.state == DeviceState.SOCFULL and d.exports_bypass:
+                        self.discharge_bypass += min(-d.pwr_produced, home)
                     self.discharge_limit += d.fuseGrp.discharge_limit(d)
                     self.discharge_optimal += d.discharge_optimal
                     self.discharge_produced -= d.pwr_produced
@@ -460,13 +465,13 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         self.power.update_value(power)
         self.availableKwh.update_value(availableKwh)
 
-        # discharge_bypass accumulates the solar-only power produced by SOCFULL devices.
-        # Subtract it from setpoint to avoid over-discharging from grid, but clamp so
-        # setpoint never goes below 0 when p1 >= 0: a SOCFULL device producing solar
-        # should still cover home demand, not trigger charge mode (fixes #1151 output
-        # cycling to 0W with bypass forbidden + 100% SoC).
-        if self.discharge_bypass > 0:
-            setpoint = max(0 if p1 >= 0 else setpoint - self.discharge_bypass, setpoint - self.discharge_bypass)
+        # Bypass production of SOCFULL devices is non-dispatchable: it keeps flowing
+        # to the home regardless of the distribution (power_charge skips devices with
+        # byPass > 0). Remove it from the dispatchable setpoint. Because the per-device
+        # bypass is capped at its homeOutput contribution, this subtraction can never
+        # push the setpoint below "p1 - real charge credits": with no device charging
+        # and p1 >= 0, the result stays >= 0 — the #1151 guarantee holds structurally.
+        setpoint -= self.discharge_bypass
 
         # Update power distribution.
         _LOGGER.info("P1 ======> p1:%s isFast:%s, setpoint:%sW stored:%sW", p1, isFast, setpoint, self.produced)


nicoh88:
I had the same problem here.

I have a SolarFlow 800 Pro and two AB2000X with 4 PV modules.
I have a SolarFlow 800 Plus with one AB2000 and one AB2000X — AC charging only.

When the SolarFlow 800 Pro was SOCFULL, but the SolarFlow 800 Plus was not, charging usually stopped automatically after a few seconds.

```
2026-06-09 18:31:03.768 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-63 isFast:True, setpoint:-274W stored:72W
2026-06-09 18:31:03.768 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -274W
2026-06-09 18:31:03.768 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -274
2026-06-09 18:31:08.773 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-76 isFast:False, setpoint:-328W stored:106W
2026-06-09 18:31:08.773 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -328W
2026-06-09 18:31:08.774 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -328
2026-06-09 18:31:13.765 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-12 isFast:True, setpoint:-318W stored:106W
2026-06-09 18:31:13.765 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -318W
2026-06-09 18:31:13.765 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -318
2026-06-09 18:31:18.764 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-21 isFast:False, setpoint:-338W stored:107W
2026-06-09 18:31:18.764 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -338W
2026-06-09 18:31:18.764 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -338
2026-06-09 18:31:23.766 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-17 isFast:False, setpoint:-349W stored:110W
2026-06-09 18:31:23.766 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -349W
2026-06-09 18:31:23.766 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -349
2026-06-09 18:31:28.764 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:12 isFast:False, setpoint:0W stored:110W
2026-06-09 18:31:28.764 INFO (MainThread) [custom_components.zendure_ha.manager] Discharge => setpoint 0W
2026-06-09 18:31:28.765 INFO (MainThread) [custom_components.zendure_ha.device] Power discharge SolarFlow 800 Plus => 0
2026-06-09 18:31:28.818 INFO (MainThread) [custom_components.zendure_ha.device] Power discharge SolarFlow 800 Pro => 0
2026-06-09 18:31:31.769 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-329 isFast:True, setpoint:-329W stored:109W
2026-06-09 18:31:31.769 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -329W
2026-06-09 18:31:36.767 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-334 isFast:False, setpoint:-334W stored:110W
2026-06-09 18:31:36.767 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -334W
2026-06-09 18:31:36.767 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -50
2026-06-09 18:31:41.787 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-344 isFast:False, setpoint:-344W stored:109W
2026-06-09 18:31:41.787 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -344W
2026-06-09 18:31:41.788 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -50
2026-06-09 18:31:46.771 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-343 isFast:False, setpoint:-343W stored:109W
2026-06-09 18:31:46.771 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -343W
2026-06-09 18:31:46.771 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -50
2026-06-09 18:31:51.766 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-352 isFast:False, setpoint:-352W stored:110W
2026-06-09 18:31:51.766 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -352W
2026-06-09 18:31:51.766 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -50
2026-06-09 18:31:56.772 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-304 isFast:False, setpoint:-353W stored:109W
2026-06-09 18:31:56.772 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -353W
2026-06-09 18:31:56.772 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -353
2026-06-09 18:31:59.766 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-223 isFast:True, setpoint:-355W stored:110W
2026-06-09 18:31:59.766 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -355W
2026-06-09 18:31:59.766 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -355
2026-06-09 18:32:03.166 DEBUG (MainThread) [custom_components.zendure_ha.manager] Update device: SolarFlow 800 Pro (6G41YYhE)
2026-06-09 18:32:03.166 DEBUG (MainThread) [custom_components.zendure_ha.manager] Update device: SolarFlow 800 Plus (F8Yp706E)
2026-06-09 18:32:03.166 DEBUG (MainThread) [custom_components.zendure_ha.manager] Finished fetching Zendure Manager data in 0.000 seconds (success: True)
2026-06-09 18:32:04.765 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-17 isFast:False, setpoint:-347W stored:110W
2026-06-09 18:32:04.765 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -347W
2026-06-09 18:32:04.766 INFO (MainThread) [custom_components.zendure_ha.device] Power charge SolarFlow 800 Plus => -347
2026-06-09 18:32:09.768 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:13 isFast:False, setpoint:0W stored:111W
2026-06-09 18:32:09.768 INFO (MainThread) [custom_components.zendure_ha.manager] Discharge => setpoint 0W
2026-06-09 18:32:09.768 INFO (MainThread) [custom_components.zendure_ha.device] Power discharge SolarFlow 800 Plus => 0
2026-06-09 18:32:09.822 INFO (MainThread) [custom_components.zendure_ha.device] Power discharge SolarFlow 800 Pro => 0
2026-06-09 18:32:12.768 INFO (MainThread) [custom_components.zendure_ha.manager] P1 ======> p1:-281 isFast:True, setpoint:-281W stored:111W
2026-06-09 18:32:12.768 INFO (MainThread) [custom_components.zendure_ha.manager] Charge => setpoint -281W
```

I manually applied the PR / the manager.py change on my setup (current GitHub master  c10b75ccd6c3ed5bd9770d6005d764202a371325 version). 

IT WORKS. ❤️ 
THANK YOU!

```
ha core stop
vi /root/homeassistant/custom_components/zendure_ha/manager.py
ha core start
```

Info for @FireSon

**BTW:**
Unfortunately, with the current GitHub master version, in my scenario:

- SolarFlow 800 Pro with 4 PV modules on the garden shed
- SolarFlow 800 Plus with AC charging only
- 2 additional PV modules with a dumb inverter on the house roof

when the SolarFlow 800 Pro is already SOCFULL, but I want to use the PV modules from the Pro to charge the SolarFlow 800 Plus, I currently have to set "Energy Export" (`select.solarflow_800_pro_grid_reverse`) to "Allow" (`allow`).

With SOCFULL and `grid_reverse` set to `forbidden`, it completely shuts off the PV from the Pro, even though the 800 Plus is not fully charged yet and is only charging from the few watts of surplus from the house roof PV.

Do you know if this is intended behavior? I may open a separate ticket for this in the next few days.

zoic21:
> BTW:
> Unfortunately, with the current GitHub master version, in my scenario:
> 
> SolarFlow 800 Pro with 4 PV modules on the garden shed
> SolarFlow 800 Plus with AC charging only
> 2 additional PV modules with a dumb inverter on the house roof
> when the SolarFlow 800 Pro is already SOCFULL, but I want to use the PV modules from the Pro to charge the SolarFlow 800 Plus, I currently have to set "Energy Export" (select.solarflow_800_pro_grid_reverse) to "Allow" (allow).
> 
> With SOCFULL and grid_reverse set to forbidden, it completely shuts off the PV from the Pro, even though the 800 Plus is not fully charged yet and is only charging from the few watts of surplus from the house roof PV.
> 
> Do you know if this is intended behavior? I may open a separate ticket for this in the next few days.

Yes a new issue will be better but I think exactly the intended behavior. For me the issue it's it's not possible to modulate solar by pass. So for example if your 800 pro produce 1000w, your home consume 200W and 800 Plus can only take 500 watts you will have 300 watts of export, since you can not force 800 pro to limit it's output to 700watts it's a problem....

nicoh88:
@zoic21 

New issue after your `manager.py` change:
https://github.com/Zendure/Zendure-HA/pull/1422/changes

When all Zendure devices are at `SOCFULL`, the house starts consuming power again, so `P1` becomes positive, and the SF800 Pro switches to solar bypass. After that, it never returns to `0` and stays in solar bypass regardless of what `P1` reports.

I hope you can follow my explanation and reproduce the issue. I will keep investigating and try to narrow it down further, and I will provide a meaningful debug log as soon as possible.

**EDIT:** Debug log added, begins 14:24:27.
[home-assistant_zendure_ha_2026-06-10T12-31-56.019Z.log](https://github.com/user-attachments/files/28794929/home-assistant_zendure_ha_2026-06-10T12-31-56.019Z.log)

**EDIT2:** Strange, it has normalized again now. However, the solar bypass never went back to `0 W`; the minimum I saw was around `30 W`, even though `0 W` would also have been fine.

zoic21:
Thanks for the log. I checked every cycle in it: `setpoint = p1 + production` holds exactly on all lines, which means `discharge_bypass` was 0 throughout — the device never matched `state == SOCFULL and exports_bypass`, so both code paths touched by this PR were no-ops and the behavior is byte-identical to master before the change.

What keeps the bypass ≥30 W is the pre-existing `byPass > 0` skip in `power_charge` (the manager can't command a bypassing device down while in charge mode), combined with the device's firmware bypass following its solar production. 

To confirm on your side: could you check, during such an episode, the Pro's `device state/socLimit`, its pass sensor, and whether energy export is set to forbidden in the Zendure app (`gridReverse`)? And if you can, revert to the commit before this PR — you should see the exact same pattern."

nicoh88:
I’m not sure. I can’t reproduce it right now because we don’t have enough sun. I’ll keep monitoring it and report back if anything does not work as expected.

zoic21:
> I’m not sure. I can’t reproduce it right now because we don’t have enough sun. I’ll keep monitoring it and report back if anything does not work as expected.

Did you have history of electricLevel, socSet  and pass between 14:24 and 14:31 ? With that I can be sure of my explication

nicoh88:
> > I’m not sure. I can’t reproduce it right now because we don’t have enough sun. I’ll keep monitoring it and report back if anything does not work as expected.
> 
> Did you have history of electricLevel, socSet and pass between 14:24 and 14:31 ? With that I can be sure of my explication

electricLevel and socSet from both pro and plus was 100.

Screenshot 2026-06-10 152422

[history.csv](https://github.com/user-attachments/files/28797776/history.csv)

zoic21:
Thanks, this history is very useful.

During the 14:24:42 to 14:25:10 window, the Pro bypass sensor is `on`, the solar passthrough is visible at around 170 W, and the manager logs show that stored increases accordingly. However, the arithmetic in the logs shows that discharge_bypass remains 0 during those cycles.

This means the condition: `state == SOCFULL and exports_bypass` was false while the bypass was active.

Another thing the manager does bring the bypass output back down. At 14:25:08, the manager sends a very low setpoint: `=> 6 W`. Immediately after that, the Pro output drops back close to zero, and the bypass sensor switches back to Aus around 14:25:10.

To fully close this, I need two additional pieces of data for the same time window:
- electricLevel and socSet for the Pro, to confirm whether the SOC was below the target while bypass was still active.
- value of the Pro Energy Export / gridReverse select.

Another thing manager no react immediatly for all change there is some timming (few second for new setpoint, and 1 minute for passing from state to another one (charge, discharge, bypass...).

nicoh88:
image

[history.csv](https://github.com/user-attachments/files/28797763/history.csv)

zoic21:
Ok its confirm my code has no impact on your issue your have `select.solarflow_800_pro_grid_reverse,forbidden`  so here : 
```
case "gridReverse":
    self.exports_bypass = value != 2
```
Your self.exports_bypass is always false (2 is forbidden, your value is never different then 2 so is always false).

On my code I just touch that : 

```
if d.state == DeviceState.SOCFULL and d.exports_bypass:
                        self.discharge_bypass += min(-d.pwr_produced, home)
```

Which is never true for you since `d.exports_bypass ` is always false.

So your issue is not due to my change it's something else.

igosoft:
Hello @zoic21 

Is self.discharge_bypass needed? I've been testing this change for some time now 

https://github.com/igosoft/Zendure-HA/pull/4/changes

zoic21:
Thanks @igosoft — you're right, `discharge_bypass` isn't strictly required. The accumulate-then-subtract is just a roundabout way of saying "don't let a SOCFULL device's bypassed solar count toward the dispatchable setpoint", which your version does more directly by skipping `setpoint += home` for those devices.

Working it through, the two are equivalent except in one case. My per-device term nets out to `max(0, homeOutput - solar) (with solar = -pwr_produced)`, while yours always nets to 0:
- For a pure solar pass-through (`homeOutput ≈ solar`) — the normal SOCFULL bypass case — both give 0, behave identically, and both kill the #1151 phantom-negative. The sensor-skew case (`solar > homeOutput`) is also handled the same way by both.
- They only diverge when `homeOutput > solar`, i.e. the SOCFULL device is briefly sourcing part of its home output from the battery. I keep that residual in the setpoint as real demand; you drop it.

Since `power_discharge` already caps SOCFULL devices to solar-only, that residual would just get redistributed onto the other devices in my version, which is arguably wrong in that transient — so I agree dropping it (your approach) is cleaner, not only simpler. I'm inclined to remove the variable and guard the credit the way you did.

One thing I'd want to keep from your branch: the `if d.state == SOCFULL and d.exports_bypass: continue` in power_charge. My PR only skips on the live `byPass > 0` sensor, so your guard also covers the window before that sensor flips — that might be behind the secondary "bypass stays elevated" behaviour nicoh88 saw.
Could you confirm I'm reading the intent right? And is the heavy `_LOGGER.info` instrumentation meant to stay or was it just for your testing?

@FireSon  Do you have an opinion on this?

igosoft:
@zoic21 you are right, the if-condition is too risky and byPass > 0 is good enough. 
Yes, the extra logs are just to confirm that the changes work (they will be reduced) . Please let me know if you find any of them useful so we can keep them

Issue #1417, opened 2026-06-08
[Bug] SolarFlow 2400 AC+ is missing plenty of useful sensors


bistory:
### Home Assistant Version

2026.6.1

### Zendure Integration Version

v1.3.1

### Describe the bug

I recently installed a SolarFlow 2400AC+, I already had a SolarFlow 2400 AC so I added the new one in HA.
However, I'm not able to add it to the energy dashboard because it's missing simple sensors like "Total Energy Charged" or "Total Energy Discharged".

Here are the sensors of 2400 AC :
<img width="176" height="794" alt="Image" src="https://github.com/user-attachments/assets/e23c1e29-4c1d-4b70-beda-233e280470ed" />

And the sensors of 2400 AC+ : 
<img width="169" height="828" alt="Image" src="https://github.com/user-attachments/assets/d2ebacf2-0dbf-4ed6-84df-d1bda6b81485" />

### To Reproduce

1. Add a 2400 AC+ to the system
2. Browse the sensors

### Expected Behaviour

We need at least "Total Energy Charged" or "Total Energy Discharged" sensors.

### What device are you using?

SF2400 AC+

### Log Extracts

```shell

```

zoic21:
Hello,

It's weird I have lot of more entities and your missing entites : 

<img width="276" height="401" alt="Image" src="https://github.com/user-attachments/assets/ac307519-f9f4-4eef-bf01-af23fbee2c6b" />

Did you try to remove the created device and redo all process of creation ?

bistory:
Yes, I did that but it didn't help...

Here are my devices : 

<img width="1020" height="533" alt="Image" src="https://github.com/user-attachments/assets/865321de-ba7c-4a8f-af3f-c0d031df1e35" />

What I did is : 
- add the battery to my system thanks to the mobile app
- reload Zendure-HA integration, then the new devices are added automatically
- it's missing a lot of sensors 🥲

zoic21:
Indeed : 

<img width="767" height="172" alt="Image" src="https://github.com/user-attachments/assets/b7fb5cd0-bb1d-4af0-8060-cb4b52ab1969" />

Have you any error on logs ?

bistory:
> Indeed :
[...]

Issue #1406, opened 2026-05-27
[Bug] Duplicate unique_id collision between SF2400 AC and SF2400 AC+ devices after upgrade to 1.3.x


antoineguilbert:
### Home Assistant Version

2026.5.5

### Zendure Integration Version

v1.3.1

### Describe the bug

When one SF2400 AC and one SF2400 AC+ are configured together, upgrading from 1.2.7 to 1.3.x causes a unique_id collision that prevents one device from registering any individual entities.

In v1.2.7, each device uses its serial suffix in the unique_id (e.g. sf2400ac_635_ac_mode for the AC, sf2400ac_856_ac_mode for the AC+), correctly disambiguating the two devices.

In v1.3.x, both the SF2400 AC and SF2400 AC+ generate identical unique_ids using the base name solarflow_2400_ac_* — the "+" distinction is lost. The first device to register claims all 76 entity slots; the second device fails to create any of its 18+ overlapping entities.

On each HA restart, 18 errors are logged:
  Platform zendure_ha does not generate unique IDs. ID solarflow_2400_ac_<entity> already exists - ignoring <entity>

Restarting HA does not resolve the issue.

### To Reproduce

Have one SF2400 AC and one SF2400 AC+ registered in the same Zendure account
2. Upgrade Zendure-HA integration from v1.2.7 to v1.3.x
3. Observe that one device loses all individual entities in Home Assistant
4. Check Settings > System > Logs: 18 "does not generate unique IDs" errors on every HA restart
5. Restart HA — errors persist

### Expected Behaviour

Each physical device should have its own distinct set of entities, disambiguated by serial number or device ID, as was correctly implemented in v1.2.7 with the sf2400ac_856_* / sf2400ac_635_* naming scheme. 
The SF2400 AC and SF2400 AC+ should not share the same unique_id prefix.

### What device are you using?

SF2400 AC+

### Log Extracts

```shell

```

### Other Information

Confirmed via entity registry inspection:
- v1.2.7: sf2400ac_635_* (72 entities) and sf2400ac_856_* (73 entities) — no collision
- v1.3.1: solarflow_2400_ac_* (76 entities, one device wins) — other device has 0 individual entities

The 18 conflicting unique_ids:
  sensors: bat_cal_time, rssi, local_apienable, ai_state, is_error
  selects: fuse_group, ac_mode, connection, grid_reverse, grid_off_mode
  switches: lamp_switch
[...]

Issue #1239, opened 2026-03-30
[Bug] 1.2.7-pre4 duplicate entity problem


thib3113:
### Home Assistant Version

2026.3.4

### Zendure Integration Version

1.2.7-pre4

### Describe the bug

If you own a SF2400AC+ and an SF2400AC => Home assistant will sanitize the entity id, by removing the + at the end . 

So, the ids will be the same !

### To Reproduce

- got two devices with nearly the same name 

### Expected Behaviour

integration will generate unique_id

### What device are you using?

SF2400 AC

### Log Extracts

```shell
1774868936688	2026-03-30T11:08:56.688Z	2026-03-30 13:08:56.676 ERROR (MainThread) [homeassistant.components.sensor] Platform zendure_ha does not generate unique IDs. ID solarflow_2400_ac_aggr_grid_off_power already exists - ignoring sensor.solarflow_2400_ac_aggr_grid_off_power
1774868936687	2026-03-30T11:08:56.687Z	2026-03-30 13:08:56.676 ERROR (MainThread) [homeassistant.components.sensor] Platform zendure_ha does not generate unique IDs. ID solarflow_2400_ac_grid_off_power already exists - ignoring sensor.solarflow_2400_ac_grid_off_power
1774868936687	2026-03-30T11:08:56.687Z	2026-03-30 13:08:56.676 ERROR (MainThread) [homeassistant.components.select] Platform zendure_ha does not generate unique IDs. ID solarflow_2400_ac_connection already exists - ignoring select.solarflow_2400_ac_connection
1774868936686	2026-03-30T11:08:56.686Z	2026-03-30 13:08:56.675 ERROR (MainThread) [homeassistant.components.sensor] Platform zendure_ha does not generate unique IDs. ID solarflow_2400_ac_switch_count already exists - ignoring sensor.solarflow_2400_ac_switch_count
1774868936686	2026-03-30T11:08:56.686Z	2026-03-30 13:08:56.675 ERROR (MainThread) [homeassistant.components.sensor] Platform zendure_ha does not generate unique IDs. ID solarflow_2400_ac_aggr_solar already exists - ignoring sensor.solarflow_2400_ac_aggr_solar
[...]

PR #1423, opened 2026-06-11
Fix unique id entity collision

zoic21:
## Problem

Entity `unique_id`s are derived from the device name via `snakecase()`, which strips non-alphanumeric characters. As a result, "SolarFlow 2400 AC" and "SolarFlow 2400 AC+" both slugify to `solarflow_2400_ac`, producing identical unique_ids. When a user owns both devices:

- HA rejects the second device's entities (duplicate unique_id), so they are never created
- `check_entities()` computes the same canonical entity_id for both devices and crashes
  with `ValueError: Entity with this ID is already registered` (reproduced in tests)

## Fix

- `unique_id` is now based on the device **serial number** (stable and guaranteed unique),  via a new `entity_unique_id()` helper shared by entity creation and migration.   Devices without a serial (Zendure Manager) keep the name-based scheme.
- The suggested `entity_id` keeps the historical name-based value, so nothing changes  visually; HA de-duplicates with a `_2` suffix when two devices share a slug.
- `check_entities()` no longer renames/removes entities owned by another device.

Fix for #1417 

## Migration — no user impact

A one-time registry migration (`Migration.async_migrate_unique_ids`, gated by config entry minor version 7 → 8) rewrites only the **invisible** `unique_id`. The `entity_id` is never touched, so automations, dashboards, scripts and recorder history keep working with zero user action. The migrated unique_id is byte-identical to what the runtime now generates, verified for every key in `translations/en.json`. The migration is idempotent and collision-safe (skip + warning, never deletes).

Also fixes `async_migrate_entry` downgrading entries to minor version 5 while the flow is at 7, which re-triggered migration on every restart.

## Tests

13 tests against real HA device/entity registries , including mutation testing (reverting each part of the fix makes the relevant tests fail), the dual AC/AC+ scenario, and the full `async_migrate_entry` upgrade path from v1.7.


thib3113:
Pretty sure this is related : https://github.com/Zendure/Zendure-HA/issues/1239 . 

Why didn't you change the entity ids ? because, it will be the same problem ? they will try to get the same name ? and so, will be happenned with number (eg `_2`) (and can change if you reset integration or things like that?) .  

On my side I use a slugifying function to output something like that : 
- select.sf2400acplus_230_grid_reverse

=> `<technicalName, with + converted to work>_<last3DigitOfSN>_<entity>`

I tried to keep the name "short", and "readable", also, I think that owning two of the same device, with the same last 3 digits will be really improbable .

( but, I agree it will need migration from users ... and I think an automated migration is not a good idea, but for the last time? )

zoic21:
> Pretty sure this is related : #1239 .
> 
> Why didn't you change the entity ids ? because, it will be the same problem ? they will try to get the same name ? and so, will be happenned with number (eg `_2`) (and can change if you reset integration or things like that?) .
> 
> On my side I use a slugifying function to output something like that :
> 
> * select.sf2400acplus_230_grid_reverse
> 
> => <technicalName, with + converted to work>
> 
> I tried to keep the name "short", and "readable", also, I think that owning two of the same device, with the same last 3 digits will be really improbable .
> 
> ( but, I agree it will need migration from users ... and I think an automated migration is not a good idea, but for the last time? )

Hello,

In my PR I move to unique_id, entity_id (show in ui) is not use anymore. Unique id is base on serial id of device so no risk of collision.

With an unique id user is able to change if he want the entity id (it's does not impact integration) and at the creation if their is already an entity id and it's not the same unique id home assistant will add _2 automatically.

If we change entity id name all user will need to redo dashboard and automation it's not possible and HA recommand to use unique id en leave entity id simple and editable by user.

thib3113:
I understand . but I'm not a big fan of this ... yes, it will fix the problem (thank you), but I think you will get problems later ... because the devices and entities are automatically created (from api), what protect your entity id ? what if you need to reinstall the integration ? ( the _2 will always keep the _2 ? or it can inverse with the other one ) . 

I understand your point, and I understand changing entity id will be a problem... but I think it's something to do as soon as possible

zoic21:
Thanks for the thoughtful feedback, I get your concern.

The key point with this PR is that by moving to unique_id, the user now has full control over the entity_id. So yes, on a reinstall (which honestly isn't something we do every day), the _2 could end up swapping between the two devices, there's no guarantee it stays attached to the same one. But the big difference is that the user can now simply rename the entity back to whatever they want. It takes a few minutes, far less painful than rebuilding every dashboard and automation, which is what would happen if we changed the entity_id itself.

I fully agree this isn't the perfect solution. The ideal would be unique_id and entity_id both based on the serial number (while being careful not to end up with ridiculously long entity ids). But I think this PR is the best option that respects the existing setup and doesn't force a breaking migration on current users.

That said, once this PR is merged, I'd like to look into a follow-up that adds a chunk of the serial number into the entity_id for newly created entities. I just want to see this one land first before investing time in another dev that may or may not be accepted.

thib3113:
> which honestly isn't something we do every day

I agree, but in general you do it when you have a problem (so it will add another problem) . 

Also, I agree that if your PR fix the problem, it is the main thing (thank you) . 

> I just want to see this one land first before investing time in another dev that may or may not be accepted.

As I said (rereading my comment, I saw that the preview doesn't show it), I think something like `<technicalName, with + converted to word>_<last3DigitOfSN>_<entity>` 

I think user will never get two same device, with the last 3 sames digits of the serial number . 
Actually, my entity with the longer name is : `sensor.solarflow_2400_acplus_230_output_pack_power_availability`

zoic21:
Hello,

Ok I had added new entity id base on device model and serial number only for new created entity, for old one no change. I think it's better it's limit issue with migration.

Just one issue remaining it's this part https://github.com/zoic21/Zendure-HA/blob/patch-3/custom_components/zendure_ha/entity.py#L292 it's rename entity with unique id it's not necessary anymore for me all this part need to me removed.

FireSon:
Hi @zoic21, thanks for the detailed work and tests. Closing in favour of a lighter fix.

**Decision.** I want to keep device names synchronized between the Zendure app and Home Assistant. Today, renaming in the Zendure app updates the HA device-registry entry and the entity_id slug follows. That's intentional: the user-facing name in both places is the single source of truth. Serial-based unique_ids would break that link, leaving entity_ids frozen on the original name while the Zendure label moves on.

**Shipping instead.** Two lines in `snakecase()`:

```python
value = value.replace("+", "_plus")
```

Fixes the SF2400AC vs SF2400AC+ collision. Users with the Zendure-default name `"SolarFlow 2400 AC+"` will see new entities created with `..._plus_...` unique_ids; the old entries stay in the registry with their history intact but stop updating. Same effect as renaming a device in the Zendure app today, and easily cleaned up by removing the orphaned entries.

Thanks again, not a rejection of the diagnosis, just a different design choice.

thib3113:
Hi @FireSon, yes it will fix the AC/AC+ collision, but not the case where you have multiple times the same product like : https://github.com/Zendure/Zendure-HA/issues/1406#issuecomment-4818915980 . 

>  That's intentional: the user-facing name in both places is the single source of truth. Serial-based unique_ids would break that link, leaving entity_ids frozen on the original name while the Zendure label moves on.

=> So, renaming the product in the zendure app will break each automation / dashboard ? ( and you will need to write a migration handling that ) . 

In my opinion the "job" of a uniq id, is to never change ... if you want to follow the zendure app, you change the name of the device, without impacting the ids ?

PR #1426, opened 2026-06-11
Add global soc, delta cellule voltage and battery efficiency

zoic21:
Hello,

This PR introduces 3 new entities:

Global SOC at the Zendure manager level
Battery voltage delta at the battery level (maxVol - minVol)
Battery efficiency aggrDischarge/aggrCharge*100
While both metrics could technically be derived using template sensors, exposing them natively improves usability and avoids unnecessary configuration on the user side.

Use cases

Global SOC
Provides a consolidated state of charge across the entire system, which is more relevant than per-battery values in most scenarios.
Battery voltage delta
Acts as a simple health indicator for individual batteries.
A sustained delta greater than ~0.15V (e.g., over 2 hours) may indicate cell imbalance and suggest that the battery requires a full charge cycle for rebalancing.
Battery efficiency
Check efficiency of battery pack


custom_components/zendure_ha/manager.py
```diff
@@ -422,6 +423,8 @@ async def powerChanged(self, p1: int, isFast: bool, time: datetime) -> None:
         availableKwh = 0
         setpoint = p1
         power = 0
+        totalStoredKwh = 0
+        onlineKwh = 0
```
igosoft:
```suggestion
        totalStoredkWh = 0
        onlinekWh = 0
```

custom_components/zendure_ha/manager.py
```diff
@@ -455,10 +458,16 @@ async def powerChanged(self, p1: int, isFast: bool, time: datetime) -> None:
...
+                totalStoredKwh += d.electricLevel.asNumber / 100 * d.kWh
+                onlineKwh += d.kWh
 
         # Update the power entities
         self.power.update_value(power)
         self.availableKwh.update_value(availableKwh)
+        if onlineKwh > 0:
+            self.globalSoc.update_value((totalStoredKwh / onlineKwh) * 100)
+        else:
+            self.globalSoc.update_value(0)
```
igosoft:
```suggestion
        self.globalSoc.update_value((totalStoredKwh / onlineKwh * 100) if onlineKwh > 0 else 0)
```

custom_components/zendure_ha/device.py
```diff
@@ -85,6 +86,20 @@ def __init__(self, hass: HomeAssistant, sn: str, parent: EntityDevice) -> None:
...
         super().__init__(hass, sn, name, model, "", sn, parent.sn)
         self.attr_device_info["serial_number"] = sn
+        self.deltaVoltage = ZendureSensor(self, "deltaVoltage", None, "V", "voltage", "measurement", 3)
+
+    def entityUpdate(self, key: Any, value: Any) -> bool:
+        """Update entity state and recalculate deltaVoltage when maxVol or minVol changes."""
+        changed = super().entityUpdate(key, value)
+        if changed and key in {"maxVol", "minVol"}:
+            max_vol = self.entities.get("maxVol")
+            min_vol = self.entities.get("minVol")
```
igosoft, line 95: just wondering, what if maxVol or minVol won't never be updated ?

custom_components/zendure_ha/device.py
```diff
@@ -163,6 +178,7 @@ def create_entities(self) -> None:
 
         self.aggrCharge = ZendureRestoreSensor(self, "aggrCharge", None, "kWh", "energy", "total_increasing", 2)
         self.aggrDischarge = ZendureRestoreSensor(self, "aggrDischarge", None, "kWh", "energy", "total_increasing", 2)
+        self.efficiency = ZendureSensor(self, "efficiency", None, "%", None, "measurement", 1)
```
igosoft: The first impression is that 'efficiency' lacks a prefix or suffix explaining what type of efficiency it refers to

custom_components/zendure_ha/device.py
```diff
@@ -222,10 +238,12 @@ def entityUpdate(self, key: Any, value: Any) -> bool:
                             self.aggrCharge.aggregate(dt_util.now(), value)
                         self.aggrDischarge.aggregate(dt_util.now(), 0)
                         self.batInOut.update_value(self.batteryOutput.asInt - self.batteryInput.asInt)
+                        self._update_efficiency()
```
igosoft: just because of current code style/structure 
```suggestion
                        self.efficiency.update_value(round(self.aggrDischarge.asNumber / charge * 100, 1) if (charge := self.aggrCharge.asNumber) > 0 else 0)
```

custom_components/zendure_ha/device.py
zoic21: I'am not sure it's possible, max/min change every time (and often) when battery charge or discharge even in idle its change. For the init it's an issue it's change when integration load so we will get an initial value without issue

custom_components/zendure_ha/manager.py
zoic21: I don't understand what change here...

custom_components/zendure_ha/device.py
zoic21: Indeed, maybe roundtripEfficiency will be better ?

custom_components/zendure_ha/manager.py
igosoft: The postfix in variable name should be kWh 
https://en.wikipedia.org/wiki/Kilowatt-hour

custom_components/zendure_ha/device.py
igosoft: sounds good, but a comment next to variable will build a context event better

custom_components/zendure_ha/device.py
igosoft: ok, I see

custom_components/zendure_ha/device.py
zoic21: Perfect it's changed.

custom_components/zendure_ha/device.py
```diff
@@ -222,10 +239,12 @@ def entityUpdate(self, key: Any, value: Any) -> bool:
                             self.aggrCharge.aggregate(dt_util.now(), value)
                         self.aggrDischarge.aggregate(dt_util.now(), 0)
                         self.batInOut.update_value(self.batteryOutput.asInt - self.batteryInput.asInt)
+                        self.roundtripEfficiency.update_value(round(self.aggrDischarge.asNumber / charge * 100, 1) if (charge := self.aggrCharge.asNumber) > 0 else 0)
                     case "packInputPower":
                         self.aggrCharge.aggregate(dt_util.now(), 0)
                         self.aggrDischarge.aggregate(dt_util.now(), value)
                         self.batInOut.update_value(self.batteryOutput.asInt - self.batteryInput.asInt)
+                        self._update_efficiency()
```
igosoft:
```suggestion
                        self.roundtripEfficiency.update_value(round(self.aggrDischarge.asNumber / charge * 100, 1) if (charge := self.aggrCharge.asNumber) > 0 else 0)
```

igosoft: I know that it looks like code duplication but it's giving a clear statement on the code block - just update values

custom_components/zendure_ha/device.py
```diff
@@ -254,6 +273,9 @@ def entityUpdate(self, key: Any, value: Any) -> bool:
 
         return changed
 
+    def _update_efficiency(self) -> None:
```
igosoft: and then this method isn't needed

igosoft (approved):

PR #1435, opened 2026-06-16
Avoid blocking file read when checking entities

zoic21:
## Problem

`EntityDevice.check_entities()` runs during integration setup (a sync method on the event loop) and lazily builds its `translation_key -> domain` map by reading `translations/en.json` from disk. That `read_text()` is blocking I/O inside the event loop and triggers Home Assistant's "Detected blocking call to open/read inside the event loop" warning at startup.

## Fix

Build the map incrementally instead of reading the file. Every entity already knows its own domain, so it records `translation_key -> domain` via `setdefault` when it is created. `checkEntity` becomes a plain class-level dict, the `json`/`pathlib` imports are dropped, and `check_entities()` no longer touches the filesystem.

The registry cleanup keeps working: the canonical `entity_id` is derived from each registry entry's own domain, not from the map — the map is only used to filter out cross-domain mismatches. The filter is relaxed accordingly so entries whose key isn't (yet) in the map are still considered.

This is the `entity.py` portion only of the original fix by @zaubara; the unrelated `POWER_START` change is left out as it is already in `master`.

### DIFF
--- a/custom_components/zendure_ha/__init__.py
+++ b/custom_components/zendure_ha/__init__.py
@@ -9,6 +9,7 @@ from homeassistant.helpers import device_registry as dr
 from .api import Api
 from .const import CONF_MQTTLOG, CONF_P1METER, CONF_SIM
 from .device import ZendureDevice
+from .entity import EntityDevice
 from .manager import ZendureConfigEntry, ZendureManager
 from .migration import Migration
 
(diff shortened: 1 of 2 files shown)


FireSon:
Hi @zoic21, thanks for splitting this out cleanly.

Two things:

**Timing.** The blocking-call warning is startup-only, not on the runtime path. For the upcoming release I want to keep changes narrow, so I'd like to hold this for the next cycle.

**Behavior shift.** This isn't purely a refactor of the map source. The filter in `check_entities()` is also relaxed from "must match" to "match-or-unknown". Combined with the loss of the `en.json` seed, the map now only contains keys for entities actually created this run. Registry entries with keys removed from the code used to be filtered out and dropped from the cleanup pass; now they survive as orphans. That silently weakens the self-healing sweep.

A possible middle ground for the next cycle: keep the incremental `setdefault` source as you have it, but keep the filter strict. The tradeoff is that any entity whose key hasn't been instantiated yet at the time `check_entities()` runs would have its registry entries wrongly removed, so we'd want to think about when to run the sweep.

Thanks again, happy to revisit after the release.

zoic21:
Thanks @FireSon, both points are fair — I agree, let's hold this for the next cycle rather than rush it in. I'll revisit the "match-or-unknown" behavior shift you flagged before resubmitting.

PR #1450, opened 2026-06-22
Modify German translations for pass_mode and grid_reverse

Kieft-C:
Updated German translations for pass_mode and grid_reverse names.

When it is right, for all other translations is also needed to do.

It seems that "pass_mode" is the old version of "grid reverse" but with same states.

### DIFF
--- a/custom_components/zendure_ha/translations/de.json
+++ b/custom_components/zendure_ha/translations/de.json
@@ -416,7 +416,7 @@
         }
       },
       "pass_mode": {
-        "name": "Bypass-Modus",
+        "name": "Energie Export old", #only at HUB1200/200 and AIO2400
         "state": {
           "auto": "Automatisch",
           "off": "Immer aus",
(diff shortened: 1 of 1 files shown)


z-master42:
Pass Mode is not the same as Grid Reverse. And why change that? For those who own a hub or an AIO, there’s nothing outdated about it, and the others shouldn’t see these options at all. And nobody who has a hub is thinking about power export anyway, and the translations are exactly as they appear in the Zendure app.

Kieft-C:
It seems that it is Not Right.

And I not change anything for HUB or AIO Generation Models.
Only the order I sort at this Point.

Old Generation Models got
"Pass_mode"

Newer Gen Models got no pass_mode and have grid_reverse
And It seems that there the Translation und Workflow was wrong.

That's why I Push the two PR #1450 and #1451

z-master42:
The translation file is for all devices. So why change Bypass-Modus to Energie Export old?

Kieft-C:
> The translation file is for all devices. So why change Bypass-Modus to Energie Export old? 

A now I understood. That I change. Okay, I will change it back. That I misunderstand in your first comment.

z-master42:
Therefore, there is no need to rename the passMode states #1451 internally either.

PR #1451, opened 2026-06-22
Update gridReverse and passMode options

Kieft-C:
parallel to PR 1450

https://github.com/Zendure/Zendure-HA/pull/1450


Kieft-C:
After testing #1450 & #1451 local at my HA Hyper System, it seems to be working, but get a strange behavior with state "auto". before i change the code, I reach state 0, 1 and 2 visible in mqtt explorer. but with no full description why.

Now with the new changes, always select "auto" jump to "forbidden" without doin anything.
select "forbidden" and "allow" stays and the feedback in mqtt explorer and app is similar, so it feels right.

But why I can't reach state "auto" / 0 anymore?
Not logic for me.

z-master42:
Why change that?

Kieft-C:
> Why change that?

See at #1450 the answer. 🤝

PR #1454, opened 2026-06-23
Update entity.py

Kieft-C:
"gridReverse" is the entity for newer gen models for export energy selection. In older models it is named "pass_mode", it seems to be same behavior and wrong in actual integration version named.

### DIFF
--- a/custom_components/zendure_ha/entity.py
+++ b/custom_components/zendure_ha/entity.py
@@ -150,7 +150,7 @@ class EntityDevice:
         "localState": ("binary"),
         "ctOff": ("binary"),
         "lampSwitch": ("switch"),
-        "gridReverse": ("select", {0: "disabled", 1: "allow", 2: "forbidden"}),
+        "gridReverse": ("select", {0: "auto", 1: "on", 2: "off"}),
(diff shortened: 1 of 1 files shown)


Kieft-C:
After testing https://github.com/Zendure/Zendure-HA/pull/1450 & https://github.com/Zendure/Zendure-HA/pull/1451 local at my HA Hyper System, it seems to be working, but get a strange behavior with state "auto". before i change the code, I reach state 0, 1 and 2 visible in mqtt explorer. but with no full description why.

Now with the new changes, always select "auto" jump to "forbidden" without doin anything.
select "forbidden" and "allow" stays and the feedback in mqtt explorer and app is similar, so it feels right.

But why I can't reach state "auto" / 0 anymore?
Not logic for me.

Now I start again with PR #1453 & #1454 - without change the "pass_mode" because it should not change.
Plus copy the state names from "Pass_mode" to "grid_reverse" because it seems, like mentioned only a generation depending entity but no difference really in between them.

May with the PR#1453 & #1454 the automatic switching from the selection will disappear.
But can't test it now in the next days.

Issue #1461, opened 2026-06-29
[Feature] Reduced idle state time during fluctuating power periods without addional wear.


mggevaer:
### Describe the feature

I noticed that Zendure Manager in smart matching mode always tries to match the grid as closely as possible, with excess wear being avoided by staying in idle if the grid keeps switching from import to export.

However, in situations where the situation switches between import/export frequently, (as with smart EV chargers that are slow to match, but do eventually match), the battery will end up spending a lot of time in idle, where even only export or only import would be better.

<img width="905" height="126" alt="Image" src="https://github.com/user-attachments/assets/51c5428b-5cf6-4115-a440-c90581ac6c73" />

I think both wear and utilisation of the battery can be improved through a (relatively) simple behavioral change.

## Improved behavior solution

When the battery will not be able to switch between charge/discharge because it has already done too many switches, instead of going to idle, go to -1W or +1W charging (or whatever the minimum is), until the "switch" timeout has passed.
That way, for example, if grid switched from export to import for about 3 seconds, the battery won't spend minutes in idle before starting to charge again. Instead it would go down to 1w charging for 3 seconds.
the same would be true if there's only surplus for a small amount of time, during a mostly discharge dominated period.

## Challenges
This all does require knowing what the "timeout" logic is for the battery, unless Zendure HA itself controls this directly.

### What device is this for?

Zendure Manager

zoic21:
Hello,
Can you check with PR #1488 if it's fix your issue ?

mggevaer:
@zoic21 apologies for slow reply, I installed it today and I'm evaluating. In the evening, when I have multiple heaters/fluctuating devices on, grid also switches between import/export a lot.
So far it looks like the battery is much quicker at initiating discharging to catch this, which is expected due to the discharge over charge bias.
Where before the battery would be mostly useless, by the time it started charging/discharging, situation would change again so it would spend most of its time in idle.
I'll report back when I have testing for how it handles working alongside the EV charging.

mggevaer:
Today was unfortunately another example where tackling this issue would make a real difference.
As shown in the image below, a lot of energy was exported due to the battery spending a lot of time in idle during periods of mostly export, and occasional short periods of import.
The green line shows the average, so battery -1.341kw, essentially means half of the time spent in idle, half spent at 2.4kw charging (max).

[...]

PR #1488, opened 2026-07-08
Reduce idle during power fluctuations

zoic21:
Implements #1461: during rapid grid direction flips (e.g. a slowly ramping EV charger), the battery no longer sits fully idle in the anti-hysteresis hold-off window. Instead, devices are kept at a minimal power of ±10 W (`SmartMode.POWER_IDLE`) so they stay in their current inverter mode and can resume full power instantly, without extra mode switches.

Flip to charging: during the 60 s hold-off window, devices that were discharging are held at +10 W instead of being stopped. If the grid flips back to export before the window expires, they ramp straight back up — no idle period, no cold start, no mode switch.
Flip to discharging: charging devices are held at -10 W for up to `HOLD_TIMEOUT ` (300 s, the same horizon already used to arm the 60 s window), but only when the discharging devices can cover the setpoint on their own. If more power is needed, they are released immediately, exactly as before. Held devices never leave charge mode, so charging resumes with no idle period at all once the window expires.
Offgrid (SF 2400), SOC-full and SOC-empty devices keep the previous behavior. `POWER_IDLE ` is 10 W rather than the ±1 W suggested in the issue because it must exceed `POWER_TOLERANCE ` (5 W), otherwise the stop command at the end of a hold would be swallowed by the device-level tolerance check.
Testing
Simulated the real `powerChanged/power_charge/power_discharge` code (HA 2026.2, fake devices reproducing the device-level clamp/tolerance logic, 2 s update ticks) and compared this branch tick-by-tick against master, with 1 and 2 devices:

- **Charge → discharge flip (800 W demand)**: identical to master at every tick — the hold never delays a needed discharge, since chargers are released whenever discharging devices can't cover the setpoint.
- **Discharge → charge flip**: identical timing to master, including at the expiry of the 60 s window (a one-cycle delay found during testing is fixed by re-adding parked devices to the idle list so they restart in the same update).
- **EV-charger fluctuation (±300 W every 15 s for 5 min)**: inverter mode switches drop from 22 to 3, MQTT commands from 33 to 24, and discharge resumes instantly at full power on each flip back (master: idle → 50 W kickstart → full power ~4 s later). Mean residual grid power slightly improves (160 W → 150 W).
- Cost of the trade-off: at most ~10 W flowing during a hold window (≈0.17 Wh per 60 s window); total energy charged/discharged in the flip scenarios is unchanged vs master.

Fixes [#1461](https://github.com/zoic21/Zendure-HA/issues/1461)

Other
I think it's need to be test because leave 10W inplace of idle is beter to restart faster charge/discharge but it's consume 0,0001667 kWh (about 0,167 Wh), it's small but not 0.

### DIFF
--- a/custom_components/zendure_ha/const.py
+++ b/custom_components/zendure_ha/const.py
@@ -71,3 +71,4 @@ class SmartMode:
 
     POWER_START = 50  # Minimum Power (W) for starting a device
     POWER_TOLERANCE = 5  # Device-level power tolerance (W) before updating
+    POWER_IDLE = 10  # Minimal hold power (W) during the direction switch timeout; must exceed POWER_TOLERANCE so the stop command is not skipped
--- a/custom_components/zendure_ha/const.py
+++ b/custom_components/zendure_ha/const.py
@@ -72,3 +72,4 @@ class SmartMode:
     POWER_START = 50  # Minimum Power (W) for starting a device
     POWER_TOLERANCE = 5  # Device-level power tolerance (W) before updating
     POWER_IDLE = 10  # Minimal hold power (W) during the direction switch timeout; must exceed POWER_TOLERANCE so the stop command is not skipped
+    HOLD_TIMEOUT = 300  # Seconds after a direction flip during which another flip is considered likely
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -536,6 +536,9 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
             else:
                 # avoid gridOff device to use power from the grid
                 await d.power_discharge(0 if d.pwr_offgrid == 0 else -10)
+                # a device parked at hold power can be restarted right away
+                if d.pwr_offgrid == 0 and d.homeOutput.asInt <= SmartMode.POWER_IDLE:
+                    self.idle.append(d)
 
         self.operationstate.update_value(ManagerState.CHARGE.value if setpoint < 0 else ManagerState.IDLE.value)
 
@@ -614,6 +617,9 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
             else:
                 # SF 2400 may show more gridInputPower than offGridPower and will be recognized as charging, so set power to 10 instead of 0
                 await d.power_discharge(0 if max(0, d.pwr_offgrid) == 0 else 10)
+                # a device parked at hold power can be restarted right away
+                if max(0, d.pwr_offgrid) == 0 and d.homeInput.asInt <= SmartMode.POWER_IDLE:
+                    self.idle.append(d)
 
         # distribute discharging devices, use produced power first, before adding another device
         dev_start = max(0, setpoint - self.discharge_optimal * 2 - self.discharge_produced) if setpoint > SmartMode.POWER_START else 0
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -514,21 +514,28 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         """Charge devices."""
         _LOGGER.info("Charge => setpoint %sW", setpoint)
 
-        # stop discharging devices
-        for d in self.discharge:
-            # avoid stopping bypassing devices
-            if d.byPass.asInt > 0:
-                continue
-            # avoid gridOff device to use power from the grid
-            await d.power_discharge(0 if d.pwr_offgrid == 0 else -10)
-
         # prevent hysteria
-        if self.charge_time > time:
+        if hold := self.charge_time > time:
             if self.charge_time == datetime.max:
                 self.charge_time = time + timedelta(seconds=2 if (time - self.charge_last).total_seconds() > 300 else 60)
                 self.charge_last = self.charge_time
                 self.pwr_low = 0
             setpoint = 0
+
+        # Stop discharging devices. While the switch timeout is running, keep them
+        # discharging at minimal power instead of going idle, so a temporary grid
+        # fluctuation (e.g. a slow ramping EV charger) can resume discharging
+        # immediately without an idle period or an extra mode switch (#1461).
+        for d in self.discharge:
+            # avoid stopping bypassing devices
+            if d.byPass.asInt > 0:
+                continue
+            if hold and d.pwr_offgrid == 0 and d.state != DeviceState.SOCEMPTY:
+                await d.power_discharge(SmartMode.POWER_IDLE)
+            else:
+                # avoid gridOff device to use power from the grid
+                await d.power_discharge(0 if d.pwr_offgrid == 0 else -10)
+
         self.operationstate.update_value(ManagerState.CHARGE.value if setpoint < 0 else ManagerState.IDLE.value)
 
         # distribute charging devices
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -77,6 +77,7 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         self.charge_weight = 0
 
         self.discharge: list[ZendureDevice] = []
+        self.discharge_time = datetime.min
         self.discharge_bypass = 0
         self.discharge_produced = 0
         self.discharge_limit = 0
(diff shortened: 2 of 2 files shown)


mggevaer:
I've had this code running for some time, looking back at the HA data, at no point during the last 3 days was sensor.zendure_manager_power 10W, which is the idle power consumption to reduce switching.
[history.csv](https://github.com/user-attachments/files/30230259/history.csv)

Perhaps the 10W value is too low?
I have a Solarflow 2400AC+

Also, based on the described algorithm, perhaps it would be of value to offer an option that has a bias towards charging instead of discharging.
In my case the battery is often not fully charged by the end of the day, while some energy is still exported due to the battery algorithm being biased towards discharging combined with occasional peeks during the day.
Although this might introduce issues with the battery fluctuating between 99% and 100% a lot more, which is not healthy AFAIK.

qpaz:
This is a code reading of the current diff (head `f87126e`), not a test — I have not installed this branch. Flagging one thing that I think survives the change, in case it is useful before merge.

The hold-off is still re-armed on every direction flip. In `power_charge` the arming line only changes its constant:

```diff
-                self.charge_time = time + timedelta(seconds=2 if (time - self.charge_last).total_seconds() > 300 else 60)
+                self.charge_time = time + timedelta(seconds=2 if (time - self.charge_last).total_seconds() > SmartMode.HOLD_TIMEOUT else 60)
```

and in `power_discharge` the reset to `datetime.max` stays a context line. So every discharge event still rewinds the window to a full 60 s, and `setpoint = 0` still applies while it runs. The charge side therefore has no upper bound on how long it can be held: if the excitation period is shorter than the window, it never expires. On our install that is not hypothetical: 31 measured flip-to-first-charge-command latencies across 12 episodes, median 71 s, 90 % of them between 60 and 80 s. Over 30 days of recorder data, 14 such episodes exported 2907 Wh of surplus while the battery sat idle with headroom. Details and the mechanism are in my comment on #1525.

One narrower point on the new hold branch:

```
hold = time < self.discharge_time and (setpoint < SmartMode.POWER_START or setpoint <= self.discharge_limit)
```

On master `discharge_limit` is reset to 0 each cycle and only incremented for devices that were classified into `self.discharge`. For a single-device installation whose only device is currently charging, `self.discharge` is empty, so `discharge_limit` is 0 and `hold` needs `setpoint < 50 W`. With a real discharge demand the charging device is released exactly as before. That is a reading of the diff plus master, not something I measured, so please treat it as a question rather than a claim: is the single-device case in scope for the hold, or intentionally out of it?

None of this is an objection to merging. Cutting mode switches from 22 to 3 and removing the cold start is clearly an improvement, and the ±10 W trade-off looks well argued. The point is only that it makes the hold cheaper rather than bounded — a cap on total hold time, or arming once per real transition instead of once per call, would be a separate and complementary change.

zoic21:
Thanks for the detailed feedback. You’re right that this PR does not bound the total charge delay during repeated discharge/charge transitions; it only makes those transitions less costly. I think limiting that delay would be better handled in a separate PR.

For a single-battery setup, releasing the charging battery when demand exceeds 50 W is intentional, since no other battery can cover the requested discharge.

Also, your measurements were made on the original 1.4.3 behavior. Testing this branch would help quantify the remaining impact.

qpaz:
Thanks — that clears up both points. Agreed that bounding the total delay belongs in its own PR, and good to know that releasing the charging battery above 50 W is intentional for a single-battery setup; that answers my question.

Happy to test, with one honest caveat on timing: the effect only shows up when there is real surplus and the battery is well below full, so I need a suitable day rather than a slot I can pick. I cannot promise a date, but I will run it and report with the same metrics as in my comment on #1525 — flip-to-first-charge-command latency, idle plateaus, and the Wh balance over the window — so the numbers are directly comparable to the 1.4.3 baseline. Excitation would be the same thermostat-cycling ~2.2 kW load as in the original report, so it is like-for-like.

Could you rebase the branch before I do? It is currently 3 ahead / 14 behind master and `mergeable: false`, so it needs a rebase for the merge in any case, and it means I install something that matches my released version rather than a mix.

For what it is worth, I did check whether the staleness would invalidate the comparison, and it would not: between your merge base `29140b9` and 1.4.3, `manager.py` differs by 6 lines — the SOCFULL solar pass-through and the `MATCHING_DISCHARGE` change — and the hold-off block itself is byte-identical, just shifted by two lines. So testing the branch as it stands would be defensible too. Your call; I am only asking because you need the rebase anyway.

zoic21:
Hello,
It's rebase, I will wait your return to see if it's fix the issue.
Thanks a lot for your test.

qpaz:
@zoic21 — short status so you are not waiting on me.

I am not going to get the staged A/B run done. The excitation I can reproduce is a
thermostat-cycling ~2.1 kW appliance, and a valid run needs it to coincide with at least
45 minutes of PV surplus and with me being free to set things up. Since the rebase on
27 July there have been four such sessions here: in three the house was a net importer
across the whole window (mean grid +1369 W, +3046 W and +836 W, exporting only 26 %, 0 %
and 8 % of the time), so there was no charge power to hold back in the first place. The
fourth was this morning, the only one with real surplus, and it was over after 22 minutes.
I would rather say this plainly than leave the PR parked on a promise. Please do not block
the merge on me — and note that @mggevaer has already run this branch for several days
and posted data.

Two things that may still be useful.

**The 1.4.3 behaviour is unchanged and still reproducing here.** During those 22 minutes
this morning, on stock 1.4.3: `bat_in_out` flat at 0 W from 10:09:59 to 10:19:25 — 9.5
minutes with no charge command while the meter was exporting; `input_limit` repeatedly
ramping 50 → 54 → 58 → ~700 W and being reset to 0; SoC pinned at 10 % throughout.

**The part that matters for #1525 is the separate PR you mentioned.** You wrote that
bounding the total charge delay would be better handled on its own, and I agree — that
bound, not this PR, is what would fix the case I reported. If you open it, I am happy to
review the diff, and once it is in a release I will report flip-to-first-charge-command
latencies and idle-plateau counts from normal operation, measured the same way as the
25 July 1.4.3 baseline so the numbers stay comparable.

Also worth a look for the underlying issue: in #1525 @uwegoexe-boop independently reported
the same pattern on 9 August with different hardware (SolarFlow 2400 AC+, 3 batteries,
Shelly Pro 3EM, 1.4.3) — charging collapsing to ~0 W while PV production rose, surplus
exported instead of stored, for roughly 2.5 hours, SoC dropping from 24 % to 21 %.

Mrdauphi:
I don't have much experience with formal testing, so I can't offer a proper A/B test. Just my practical experience running this for a while.

I've been running this on the rebased branch for over a week now, and I can confirm this feature works well for my situation.

I noticed that when running my oven, it doesn't draw power steadily but fluctuates during operation. This caused my batteries (2x 800+) to keep switching between discharging, idle, and charging, especially when there was enough solar surplus to charge from.

This change now keeps the batteries at a steady ~10W discharge most of the time, while still being able to ramp up quickly when the oven draws more power. So for my setup, it makes the charging/discharging behavior noticeably more stable.

mggevaer:
@Mrdauphi Perhaps I didn't notice an issue because my focus has been more on the excess surplus not being captured. While this fix might perhaps do more for preventing import when the battery is not yet empty.
However at the end of the day I often don't have a fully charged battery yet so any loss of charging is usually a loss of savings on my end, while there's plenty of time to discharge the battery fully during the night.

igotsoul:
Following up on the point that bounding the total hold-off belongs in a separate PR: that is #1663. It keeps the 60 s / 2 s / 300 s values but caps the hold-off at 60 s after the *first* arming of a flip episode, so repeated flips can no longer extend it indefinitely (with a regression test reproducing #1525).

The two changes are complementary: this PR makes the hold-off period cheaper (no idle, no extra mode switch), #1663 makes it finite. #1665 adds a third angle (no debounce in MANUAL / STORE_SOLAR). All three touch the `# prevent hysteria` block in `power_charge`, so I'm happy to rebase #1663 onto whichever lands first.

A data point for the single-battery case, since the discharge-side hold only applies below 50 W there: on a SolarFlow 2400 AC+ I'm currently testing a workaround at the P1 level that caps short import spikes so the battery keeps charging at ~50 W instead of flipping. With a kettle (~1.5–2 kW, 7–13 s) the battery stays in charge mode and is back at full charge power within ~7–8 s, instead of idling 24–60 s. It confirms that holding a minimal power in the current direction is the right idea. For single-battery setups the hold would need to cover large short spikes as well, not only < 50 W.

PR #1507, opened 2026-07-12
Send full command incl. smartMode/acMode for zenSDK limit writes

### LINKED ISSUE #1505: [Bug] SF2400 Pro stops discharging after reaching 100% SoC — output stays at 0 W despite valid output limit (zenSDK)
### Home Assistant Version

core-2026.5.4 (Container, amd64)

1.4.2 (HACS)

### Describe the bug

When the SolarFlow 2400 Pro reaches 100% SoC, it enters standby and afterwards ignores all discharge commands. number.solarflow_2400_pro_output_limit is accepted (state updates to the requested value, e.g. 2400), ac_mode is output, but output_home_power stays at 0 W indefinitely. The house then draws everything from the grid even though the battery is full and discharge is requested. Only a device restart (power cycle / app) recovers it.
Even PV would be possible but stays at 0. 

Environment

- Home Assistant: core-2026.5.4 (Container, amd64)
- Zendure-HA integration: 1.4.2 (HACS)
- Device: SolarFlow 2400 Pro (3 packs: 1× I2400, 2× AB3000L)
- Connection mode: zenSDK (local)
- Fuse group: owncircuit, grid reverse: allow

Steps to reproduce

1. Operate the SF2400 Pro in ac_mode: output with an external controller writing number.solarflow_2400_pro_output_limit (load-following).
2. Let the battery charge to 100% SoC (soc_set = 100).
3. Keep requesting discharge via output_limit > 0.
4. Observe output_home_power drop to 0 W and never recover, regardless of the requested limit.

Observed timeline (2026-07-12, times UTC)

┌────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────┐
│     Time (UTC)     │                                             Event                                              │
├────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ until 15:20        │ Device discharging normally, output_home_power ≈ 780–920 W following output_limit              │
├────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 15:20:56           │ electric_level jumps 98% → 100%                                                                │
├────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 15:21:07–15:21:46  │ Controller raises output_limit: 861 → 1411 → 1961 → 2400 W (all accepted by the number entity) │
├────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 15:21:20           │ output_home_power falls to 0 W and stays there                                                 │
├────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────┤
│ 15:21 → now (>1 h) │ solar_input_power = 0, pack_input_power = 0, output_home_power = 0, grid import ≈ 930 W        │
└────────────────────┴───────────────────────────────────────────────────────────────────┘

Relevant states while stuck

sensor.solarflow_2400_pro_electric_level        100 %
select.solarflow_2400_pro_ac_mode               output
number.solarflow_2400_pro_output_limit          2400.0
sensor.solarflow_2400_pro_output_home_power     0 W
sensor.solarflow_2400_pro_pack_input_power      0 W
sensor.solarflow_2400_pro_inverse_max_power     2400   (NOT the known 800 W issue)
sensor.solarflow_2400_pro_ac_coupling_state     57344
sensor.solarflow_2400_pro_ac_status             1
binary_sensor.solarflow_2400_pro_hems_state     off
number.solarflow_2400_pro_min_soc               5.0
number.solarflow_2400_pro_soc_set               100.0

### To Reproduce

1. Operate the SF2400 Pro in ac_mode: output with an external controller writing number.solarflow_2400_pro_output_limit (load-following).
2. Let the battery charge to 100% SoC (soc_set = 100).
3. Keep requesting discharge via output_limit > 0.
4. Observe output_home_power drop to 0 W and never recover, regardless of the requested limit.

### Expected Behaviour

After reaching 100% SoC the device should resume dischargint_limit is requested in ac_mode: output (or the integrationshould re-send/refresh whatever command is needed to wake the device from the 100% standby state).

SF2400 Pro

### Log Extracts

```shell

```

### Other Information

- Reproduced multiple times; it happens specifically after hitting 100% SoC. Workaround so far: restart the device, or keep soc_set below 100%.
- No errors or warnings from zendure_ha in the HA log whilehe only zendure line is the generic custom-integrationloader warning.
- This looks related to, but distinct from, the sporadic insue: here inverse_max_power stays at 2400 the whole time and no out_of_range occurs — the device simply does not act on the accepted output limit.
- Happy to provide diagnostics dump or debug logs (custom_con request.

krenk80:
## Problem

When an external controller (automation, energy-management add-on, …) writes `number.output_limit` or `number.input_limit`, `ZendureZenSdk.entityWrite` posts only the bare property:

```json
{"properties": {"outputLimit": 800}}
```

The integration's own control path (`discharge()` / `charge()`) always sends the full command including `smartMode` and `acMode`.

A bare limit write is **silently ignored** when the device has dropped out of smart mode. Observed on a SolarFlow 2400 Pro at 100% SoC (see #1505): the device stopped discharging (0 W output), kept reporting telemetry normally, but ignored every subsequent `outputLimit` write for over 90 minutes. Recovery was only possible by re-asserting the full command (`smartMode: 1, acMode: 2, …`) against the local API.

## Fix

Route `outputLimit` / `inputLimit` writes through `doCommand()` with the same full payload that `discharge()` / `charge()` use, so every external write re-asserts `smartMode`/`acMode`. `doCommand()` also keeps the cloud/zenSDK transport distinction intact (MQTT publish vs. local HTTP).

The `smartMode: 0` case for `value == 0` (with no off-grid power) mirrors `power_off()`.

## Testing

- Reproduced the hang on a SolarFlow 2400 Pro (FW-state after ramping `outputLimit` to 2400 W at 100% SoC): bare property writes ignored, full command accepted.
- Verified the full-command payload live against the device's local `properties/write` endpoint: discharge resumed and the device accepted the full 2400 W again.

Fixes #1505

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -745,7 +745,15 @@ class ZendureZenSdk(ZendureDevice):
             _LOGGER.error("Entity %s has no translation_key, cannot write property %s", entity.name, self.name)
             return
 
-        if self.online and self.connection.value == 0:
+        # Route limit writes through the power routines so they send the full command
+        # (smartMode/acMode), exactly like the manager does. A bare outputLimit/inputLimit
+        # property write is silently ignored when the device has dropped out of smart mode
+        # (observed on SolarFlow 2400 Pro at 100% SoC, see #1505).
+        if entity.propertyName == "outputLimit":
+            await self.discharge(value)
+        elif entity.propertyName == "inputLimit":
+            await self.charge(-value)
+        elif self.online and self.connection.value == 0:
             await super().entityWrite(entity, value)
         else:
             _LOGGER.info("Writing property %s %s => %s", self.name, entity.propertyName, value)


FireSon:
Thanks @marcokrenkel-collab — the diagnosis (bare limit writes being ignored after the device drops out of smart mode) is correct and worth fixing.

I pushed a small revision so it reuses the existing routines instead of hand-rolling the command payload: `entityWrite` now delegates `outputLimit` -> `discharge(value)` and `inputLimit` -> `charge(-value)`, with the normal property-write path as the `else`. Same full-command behaviour you intended (smartMode/acMode + both limits), but a single source of truth, and it inherits the kickstart-walk handling automatically. Also rebased onto current master.

Issue #1502, opened 2026-07-12
[Feature] SolarFlow 4000 Mix Pro: Token accepted but no devices found


stepstark:
### Describe the feature

I’m using the official Zendure integration in Home Assistant.  The token is accepted successfully, but Home Assistant shows 'No devices found for this account.' Device model is Zendure SolarFlow 4000 Mix Pro. 
The device is visible and controllable in the Zendure app. Question: Is the 4000 Mix Pro currently unsupported by this integration? If so, are there plans to add support, or is there a recommended workaround?

### What device is this for?

Hyper2000

rrozema:
Same problem for solarflow 4000 mix ac+. Unable to add the integration due to this problem.

UPS90:
Same here, it looks like the 4000MIX AC+ is not yet in the configuration.

jmguil:
Think is it still under dev. According to the code, in the directory devices there is a reference to SolarFlow4000AC+. But in device.py for retreiving the battery capacity, there is no case for 4000Mix.

luckymalte:
I have the same issue with two SolarFlow 4000 Mix Pro devices. SolarFlow 800 Plus is detected, but both 4000 Mix Pro devices are missing.

gjburg:
Same issue overhere new 4000Ac+ 

Logger: custom_components.zendure_ha.api
Bron: custom_components/zendure_ha/api.py:217
Integratie: Zendure Home Assistant Integration ([documentatie](https://github.com/zendure/zendure-ha), [problemen](https://github.com/zendure/zendure-ha/issues))
Eerst voorgekomen: 20:56:44 (1 gebeurtenis)
Laatst gelogd: 20:56:44

Zendure API does not reply any devices: {'code': 200, 'success': True, 'data': {'mqtt': {}, 'deviceList': []}, 'msg': 'Operation successful'}

gjburg:
Indeed, as workaround I've now used https://github.com/Gielz1986/Zendure-HA-zenSDK/tree/main/Global%20(EN)%20Integration which creates halpers. In the IP helper fillin the IP of the zendure battery and make sure the lease is static. 

http://<ip-van-zendure>/properties/report

Then add the energy export and import to your energy settings of HA. Which works for me for now. Hopefully there will be an official solution from zendure too.

Rolf-Smit:
[...]

PR #1528, opened 2026-07-19
Add local (cloud-free) device onboarding/configuration

Rolf-Smit:
### Summary

Adds a local (cloud-free) onboarding option to the config flow that talks directly to ZenSDK-capable devices over their on-device HTTP API, no Zendure account token needed. Requires knowing the device's IP and assigning it a static address.

This works around the "no device found" issue reported in #1502 (e.g. affects the SolarFlow 4000 Mix, which the cloud API doesn't seem to return). But can also be used by those that want a true local setup, that does not require a cloud based MQTT connection.

Devices onboarded locally behave identically to cloud-onboarded ones, same entities, same manager logic.

### Detailed changes

- **`config_flow.py`**: new cloud/local pick menu, plus an add device wizard for local devices. Each device is validated via its `/properties/report` endpoint, with the model auto-detected from the reported product string. Reconfigure supports adding/removing devices later.
- **`api.py`**: in local mode, `Connect()` synthesizes the device list from config instead of hitting the cloud API and skips cloud MQTT setup; `localDeviceList()` builds one entry per configured device; `testLocalDevice()` validates a device and returns `(sn, product)`. Also registers the `solarFlow4000MixAC+` product alias to `SolarFlow4000AC_Plus`.
- **`device.py`**: local devices default to (and are locked to) the ZenSDK connection, and poll the local report every cycle since there's no MQTT push.
- `Api.devices`: is now cleared on unload and on device removal, so a removed device doesn't reappear.

### Localization
I manually worked on Dutch and English, however an LLM was used to translate to other languages. Would be appreciated if someone could check those.

### Test plan
- [X] Add a local device via the new config flow menu and confirm it validates and creates entities
- [ ] Confirm a cloud-onboarded setup still works unchanged
- [X] Remove a local device via reconfigure and confirm it doesn't reappear on reload
- [X] Confirm the last device on a local-only entry can't be removed (only the whole entry)

Unfortunately I cannot easily test if cloud-onboarded devices/setup still works as expected, as I don't have any devices that work via that method. Perhaps someone else could validate, but the code path is basically unaffected so I don't expect any issues.

Please let me know what you think of this PR, I'm happy to work on details where required.

### DIFF
--- a/custom_components/zendure_ha/translations/fr.json
+++ b/custom_components/zendure_ha/translations/fr.json
@@ -2,7 +2,8 @@
   "config": {
     "abort": {
       "already_configured": "L'appareil est déjà configuré",
-      "reconfigure_successful": "Reconfiguration réussie"
+      "reconfigure_successful": "Reconfiguration réussie",
+      "no_devices_configured": "Aucun appareil restant. Ajoutez au moins un appareil avant d'enregistrer."
     },
     "error": {
       "cannot_connect": "Impossible de joindre l'API Zendure. Vérifiez votre connexion Internet et réessayez.",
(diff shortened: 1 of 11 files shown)


custom_components/zendure_ha/translations/fr.json
jahierwan, line 1: This translation looks fine, thanks

Rolf-Smit:
@FireSon is this something support can be added for? Its now closed as stale, but nobody really looked at it.

Issue #1536, opened 2026-07-21
[Bug] Battery A discharges while Battery B charges


Thialha:
### Home Assistant Version

2026.7.2

### Zendure Integration Version

v1.4.3

### Describe the bug

Hello,
I have two SF 800+ units, each connected to its own battery.
The two SF 800+ units are installed on separate AC circuits, each protected by its own dedicated circuit breaker.

I've noticed an issue that occurs consistently around sunrise and sunset, when solar production is low.
Instead, one battery starts discharging to supply the house, while the other battery charges from the available solar production.

In the attached example:

Solar production: 123 W
House consumption: 112 W
Battery A: discharging 23 W
Battery B: charging 21 W
Grid import/export: approximately 0 W

From what I can see, Battery A is not directly charging Battery B. Rather, Battery A is supplying the house while the available PV production is used to charge Battery B.

This results in unnecessary battery cycling. Since the PV production is already insufficient to cover the house consumption, I would expect both batteries to discharge to support the load, instead of charging one battery while the other is discharging.

The issue occurs consistently around sunrise and sunset.

See the attached screenshot illustrating this behavior.

Is this the expected operating behavior, or could this be a firmware/software issue?

Thank you.

<img width="698" height="1051" alt="Image" src="https://github.com/user-attachments/assets/0bbe4e29-754d-4598-a6b1-cd82f3ff8792" />
<img width="721" height="1016" alt="Image" src="https://github.com/user-attachments/assets/627014c3-bb29-41ce-958b-27ec4e4a8c3c" />
<img width="773" height="1113" alt="Image" src="https://github.com/user-attachments/assets/e090ef6e-420f-46b8-8593-42c968ac6711" />

<img width="1206" height="1417" alt="Image" src="https://github.com/user-attachments/assets/522a6e94-8d46-4119-8f95-48ff9c77d0db" />

### To Reproduce

Around sunrise and sunset, Battery A discharges while Battery B charges.
The power transfer is typically around 0–80 W, even though there is almost no power exchange with the grid.

### Expected Behaviour

During these periods, I would expect both batteries to discharge (or at least neither battery to charge) in order to supply the house load.

### What device are you using?

SF800 Plus

### Log Extracts

```shell

```

surfer1264:
Maybe the Same problem

![image](https://github.com/user-attachments/assets/3f350409-a701-4912-921f-3bb4986c4160)


[...]

Issue #1525, opened 2026-07-19
SF2400 AC rapidly flip-flops charge/discharge in smart mode, exporting surplus to grid


Philippe26150:
### Home Assistant Version

2026.7.2

### Zendure Integration Version

1.4.3

### Describe the bug

With Zendure Manager operation mode set to `smart`, the SolarFlow 2400 AC rapidly and repeatedly flip-flops between charging and discharging over a period of just a few seconds, even during hours of strong, stable solar production and while the battery is far from full (35-36% SoC, max SoC set to 100%). During the charge/discharge transitions, solar surplus that should be going into the battery is instead exported to the grid.

This happens throughout the middle of the day, not just at the "cloud passing" edge case — production was steady in the 3500-4600 W range during the observed windows.

Environment:
- Fuse group: owncircuit
- AC connection mode: zenSDK
- P1 meter: Shelly Pro 3EM
- Input limit (number.solarflow_2400_ac_input_limit): dynamic, observed ranging ~750-2050 W (max configured 2400 W)
- Min/Max SoC: 5% / 100%

Example window (2026-07-19, ~12:55-13:04 local time), production steady around 4500-4650 W, battery at ~24-27% SoC:

| Time | Charge power | Discharge power | Grid meter (negative = export) |
|---|---|---|---|
| 12:55:54 | 0 W | 425 W | +425 W (import) |
| 12:56:01 | — | 625 W | +575 W (import) |
| 12:56:35 | 0 W | — | -481 W (export) |
| 12:56:40 | 0 W | — | -377 W (export) |
| 12:57:02 | 0 W | — | -280 W (export) |
| 12:57:35 | 0 W | — | -299 W (export) |

Charge power drops to 0 W and discharge simultaneously spikes to 200-600+ W within the same few-second window, repeatedly, all while solar production is abundant and the battery has ample headroom to keep charging.

A second observation window after a full Home Assistant restart (which cleared unrelated TimeoutError communication issues) still showed the same pattern — charge power ramping smoothly for ~10-30 seconds, dropping to 0, and grid export spiking to as much as -1191 W during the same period, indicating the underlying oscillation is not caused by the network timeouts.

### To Reproduce

1. Set Zendure Manager operation mode to "smart" on a SolarFlow 2400 AC with a P1 meter (Shelly Pro 3EM) configured.
2. Let it run during a sunny, high-production period with battery SoC well below the configured max.
[...]

PR #1538, opened 2026-07-24
Debounce smartMode/acMode re-assertion on zenSDK limit writes

### LINKED ISSUE #1521: After Update to 1.4.3 the smart mode alternates after every mqtt command
### Home Assistant Version

2026.7.2

1.4.3

### Describe the bug

after updating in first step one of my two batteries went to Standby.
Shortly after the other did so too.
I figured out that in the activities of the zendure batteries the automation appeared and after every command to set a new input level the smart mode was switched to 0 and then again to 1 which runs over the mqtt commands and sets input and output to 0.
The activities mentioned that this was initiated by an automation, but I never used the smart mode in any automation.
Switching back to v1.4.2 solved the problem!

### To Reproduce

Install v1.4.3 and try to change input or output values of the appropriate batterie

### Expected Behaviour

-

SF2400 AC+

### Log Extracts

```shell

```

### Other Information

_No response_

krenk80:
## Problem

PR #1507 routed every `number.output_limit` / `number.input_limit` write through `discharge()`/`charge()`, which unconditionally sends the full command (`smartMode` + `acMode`) alongside the limit. That fixed the original problem — a bare property write is silently ignored once the device has dropped out of smart mode (#1505) — but it now forces a `smartMode`/`acMode` re-assertion on **every single write**, including the manager's own alternating charge/discharge dispatch and repeated writes from an external automation, even when the direction hasn't actually changed.

Reported in #1521: after updating to 1.4.3, "the smart mode alternates after every mqtt command" — `smartMode` flips 0→1→0 and `acMode` bounces input/output on every command, batteries drop to Standby. Multiple users confirmed, and one traced it to the exact commit (`f4e2267`, i.e. #1507).

## Fix

Only include `smartMode`/`acMode` in the outgoing command when:
- the requested direction (`acMode`) differs from the last one this device instance asserted, or
- `MODE_REASSERT_INTERVAL` (30s) has elapsed since the last assertion.

Otherwise only the bare `outputLimit`/`inputLimit` is sent — matching pre-1.4.3 behavior for steady-state writes. The periodic re-assertion (rather than "only on change") is kept so a silent smart-mode drop still self-heals within 30s without requiring an explicit direction change first, preserving the #1505 fix.

`charge()`, `discharge()`, and `power_off()` (used by both `entityWrite` and the manager's `power_charge`/`power_discharge` dispatch) now share this logic via a new `ZendureZenSdk._sendPower` helper, so manager-driven and externally-driven writes stay consistent about what mode was last asserted.

## Testing

- Verified `device.py`/`const.py` parse cleanly.
- Traced the call paths for `entityWrite` (external/automation writes) and `manager.power_charge`/`power_discharge` (built-in smart dispatch) to confirm both route through the same debounced `_sendPower`, so the fix applies uniformly regardless of the caller.
- No existing test suite in the repo to run against.

Fixes #1521

## Related (same 1.4.3 symptom cluster, not fixed by this PR)

- #1525 — SF2400 AC rapidly flip-flops charge/discharge in smart mode. This is the manager's own dispatch (`power_charge`/`power_discharge`) legitimately alternating direction, so the debounce here doesn't suppress it — looks like a separate hysteresis issue in `manager.py`'s dispatcher, not caused by #1507.
- #1522 / #1536 — discharge/charge behavior around low-production periods; also traced by a commenter to a possible missing dead-zone in `manager.py`'s dispatcher rather than the `entityWrite` path this PR touches.

### DIFF
--- a/custom_components/zendure_ha/const.py
+++ b/custom_components/zendure_ha/const.py
@@ -71,3 +71,5 @@ class SmartMode:
 
     POWER_START = 50  # Minimum Power (W) for starting a device
     POWER_TOLERANCE = 5  # Device-level power tolerance (W) before updating
+
+    MODE_REASSERT_INTERVAL = timedelta(seconds=30)  # Min interval between forced smartMode/acMode re-assertions on unchanged direction
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -724,6 +724,8 @@ class ZendureZenSdk(ZendureDevice):
         super().__init__(hass, deviceId, name, model, definition, parent)
         self.connection = ZendureRestoreSelect(self, "connection", {0: "cloud", 2: "zenSDK"}, self.mqttSelect, 0)
         self.httpid = 0
+        self._modeAsserted: tuple[int, int] | None = None
+        self._modeAssertedAt: datetime = datetime.min
 
     async def mqttSelect(self, select: Any, _value: Any) -> None:
         from .api import Api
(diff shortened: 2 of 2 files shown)


FireSon:
Thanks for this, and for tracking down the root cause in #1521.

I have gone a step further and reverted f4e2267 entirely on master (`6fc073d`). Two reasons.

The regression is well confirmed: four reporters, two of whom verified it by downgrading to 1.4.2. And it hits exactly the people who drive the number entities from their own automations, which is a real part of the user base.

On the other side, the bug the commit was meant to address (#1505) was never confirmed fixed after the 1.4.3 release, and the recovery you measured yourself was a different sequence than what the commit implemented: limit to 0, wait for the device to start moving on its own, limit back. That reads as device behaviour at 100% SoC rather than something the write path can repair, and I would rather keep that path plain than carry a timer inside it.

That leaves nothing here to debounce, so I am closing this for now. Happy to reopen if a better solution for #1505 turns up.

PR #1540, opened 2026-07-24
Add a pytest test environment (device/number/manager scenario tests)

### LINKED ISSUE #1539: [General] Tests should be added to cover all scenarios and ensure PR checkins do not break behavior
### Describe the feature

Tests should be added to cover all scenarios and ensure PR checkins do not break behavior

Tests should run during PR and Main build

### What device is this for?

Zendure Manager

### Other Information

_No response_

krenk80:
Fixes #1539

## Why

This repo currently has zero automated tests and no CI step that executes any Python code — `.github/workflows/` only runs HACS metadata validation and hassfest. That gap is exactly how the #1521 regression (introduced by #1507) reached 1.4.3 unnoticed: nothing exercised the `entityWrite`/`charge`/`discharge` code path before merge, and reviewers could only read the diff.

## What this adds

- `requirements_test.txt` / `pyproject.toml` — pytest config (`pytest-homeassistant-custom-component`, the standard harness for HA custom integrations; it provides a realistic `hass` with working device/entity registries)
- `tests/conftest.py` — a `zensdk_device` fixture that builds a real `ZendureZenSdk` device with `mqttPublish` captured instead of hitting the network, plus stubbing the `.add`-callback pattern entities normally get from platform setup
- `tests/test_device_zensdk.py` — the #1505 (bare writes silently ignored after a smart-mode drop) vs #1521 (mode flips on every write) scenarios, covered via #1538's debounced `_sendPower`, plus `entityWrite` routing and the charge/discharge kickstart logic
- `tests/test_number_clamp.py` — #1506's "clamp to the device's current range instead of raising out_of_range" behavior
- `tests/test_manager_dispatch.py` — `ZendureManager.power_charge`/`power_discharge` dispatch, exercised via a lightweight harness bound to the real methods (constructing a full `ZendureManager` needs a config entry, P1 meter wiring, DataUpdateCoordinator setup, etc. unrelated to the dispatch logic itself)
- `.github/workflows/test.yml` — runs pytest on push/PR on `ubuntu-latest`

## Depends on #1538

This branch is stacked on #1538 (the debounce fix) because several of the new tests assert the *fixed* smartMode/acMode debounce behavior — run against current `master` alone they correctly fail, since that's the still-open #1521 bug. Please merge #1538 first; this PR's diff will shrink down to just the test files once that lands (GitHub is showing #1538's commit here too in the meantime).

## Verified

All 19 tests pass in CI on this branch (ubuntu-latest, Python 3.14): https://github.com/krenk80/Zendure-HA/actions/runs/30081678878

Note: Home Assistant core's test tooling imports `fcntl` (POSIX-only) and `device.py`'s module-level `from homeassistant.components import bluetooth` transitively pulls in several Linux-oriented optional component dependencies (`aiousbwatcher`, `serialx`, `aioesphomeapi` — pinned in `requirements_test.txt`), so this suite runs on Linux/macOS CI, not native Windows.


FireSon:
Happy to add the pytest test environment! There is no Python test coverage in CI at all today, so a pytest setup plus a workflow that runs it on every PR is a real step forward. The conftest work to get the Home Assistant fixtures going is the awkward part, and you have already solved it.

I have trimmed the branch and force-pushed, since maintainer edits are enabled.

What changed: f4e2267 was reverted on master (`6fc073d`), so the #1538 change this branch carried is gone, along with the five scenarios asserting the mode debounce and the entityWrite routing. The other six zenSDK scenarios still describe current behaviour, so those are kept, together with the manager dispatch and number clamp tests.

Hopefully this is the start of broader coverage. The power distribution in the manager is where I would most like to see it grow next, since that is where most of the reports keep landing.

Issue #1544, opened 2026-07-25
Are outputLimit/inputLimit property writes flash-backed on zenSDK devices (e.g. SolarFlow 800 Pro / CO4A)?


Patrickt21:
While reviewing `ZendureZenSdk.charge()`/`discharge()` I noticed that direct property
writes to `outputLimit`/`inputLimit` (via `properties/write` over the local zenSDK HTTP
API) have no rate-limiting anywhere in the zenSDK code path — unlike `devices/ace1500.py`,
which explicitly quantizes and throttles its `inputLimit` writes to protect flash:

> "inputLimit writes go to the ACE 1500's flash memory. Quantize to 50 W steps and
> throttle to one write per 30s so a long surplus-tracking session doesn't burn through
> flash endurance."

Since `entityWrite()` now routes Number-entity writes for `outputLimit`/`inputLimit`
through `charge()`/`discharge()` (same as the Manager's P1-loop), a user automation that
drives `number.<device>_output_limit` directly for zero-feed-in control (common pattern,
e.g. reacting to a P1/smart-meter sensor every 1-4s) can burst-write to the device's local
HTTP server far more often than the Manager's own `SmartMode.TIMEFAST`/`TIMEZERO`
(2.2s/4s) would ever allow.

Questions:
1. Are `outputLimit`/`inputLimit` persisted to flash/NVS on zenSDK devices (SolarFlow
   800/800+/800 Pro, 1600, 2400 series, 4000), the same way they apparently are on the
   ACE 1500?
2. Is there a known safe write-rate for the local HTTP API (`properties/write`) on these
   devices, or observed issues (webserver hangs, dropped connections) under frequent
   writes?

I have a small PR ready (quantize 10W / min-interval 2s inside
`ZendureZenSdk.charge()`/`discharge()`, kept below `TIMEFAST` so it never changes Manager
behavior) if this is a real concern — happy to open it, but wanted to check the flash
assumption with you first before asserting it as fact in the PR description.

harrymayr:
Smartmode = 0 writes into flash.

<img width="2340" height="1080" alt="Image" src="https://github.com/user-attachments/assets/47d11c25-41c2-41f5-b370-e140bc6d3460" />

PR #1545, opened 2026-07-25
Throttle direct outputLimit/inputLimit writes in ZendureZenSdk

Patrickt21:
## What

Adds a small quantize (10W) + minimum-interval (2s) guard inside
`ZendureZenSdk.charge()`/`discharge()`, mirroring the pattern
`devices/ace1500.py` already uses for its `inputLimit` writes.

## Why

`entityWrite()` routes direct Number-entity writes for `outputLimit`/`inputLimit`
through `charge()`/`discharge()` — the same path the Manager's own P1-loop uses. The
Manager debounces itself via `SmartMode.TIMEFAST`/`TIMEZERO` (>=2.2s apart), but that
protection only applies when the Manager is driving the device. A user automation that
writes `number.<device>_output_limit` directly (e.g. a zero-feed-in loop reacting to a
smart-meter sensor that updates every 1-4s) has no floor at all today and can burst-write
to the device's local HTTP server many times per second.

This mirrors exactly what `ace1500.py` already does for its (flash-backed) `inputLimit`
— see #1544 for the open question of whether the same is true for zenSDK
devices' `outputLimit`/`inputLimit`. Even if it turns out these aren't flash-backed,
bounding the write rate to the device's local web server seems worth doing regardless.

## Design notes

- `POWER_PROPERTY_MIN_INTERVAL` (2s) is deliberately kept *below* `TIMEFAST` (2.2s) so it
  never changes Manager behavior in the normal case — it only catches the previously
  unprotected direct-write path.
- `power == 0` always passes through immediately, unthrottled — stop requests should
  never be delayed.
- `power_off()` bypasses `charge()`/`discharge()` entirely (calls `doCommand()` directly)
  and stays unthrottled, consistent with `ace1500.py`'s `power_off()`.
- Quantization step (10W) is close to the existing `SmartMode.POWER_TOLERANCE` (5W)
  dead-band already applied one layer up in `power_charge()`/`power_discharge()`, so it
  shouldn't meaningfully change control precision for existing users.

## Testing

- `python -m py_compile` clean.
- Manually diffed against `origin/master` (00b70f0) to confirm no interaction with the
  recent `entityWrite()` routing change (#1505 fix).
- Not tested against real hardware — I'd appreciate a maintainer or someone with a zenSDK
  device sanity-checking the interval/step choices before merge.

Happy to adjust the constants based on your answer in #1544.

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -40,6 +40,18 @@ CONST_HEADER = {"content-type": "application/json; charset=UTF-8"}
 CONST_TIMEOUT = ClientTimeout(total=4)
 SF_COMMAND_CHAR = "0000c304-0000-1000-8000-00805f9b34fb"
 
+# Quantize + rate-limit outputLimit/inputLimit writes issued via ZendureZenSdk.charge()/
+# discharge(). The Manager's own P1-loop already debounces these calls (SmartMode.TIMEFAST/
+# TIMEZERO, >=2.2s apart), but that loop is not the only caller: entityWrite() routes direct
+# Number-entity writes (e.g. a user automation driving output_limit for zero-feed-in) through
+# the same charge()/discharge() methods with no floor at all, so a noisy P1 sensor can burst
+# HTTP writes to the device's local web server many times per second. POWER_PROPERTY_MIN_INTERVAL
+# is kept below TIMEFAST so it never throttles the Manager's normal cadence - it only catches
+# the previously-unprotected direct-write path. power == 0 always passes through immediately
+# so stop requests are never delayed.
+POWER_PROPERTY_STEP_W = 10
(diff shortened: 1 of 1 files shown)


FireSon:
Thanks for this, and for the clear write-up in the comment block.

I have reverted f4e2267 on master (`6fc073d`). `entityWrite` writes a single property again, so the direct Number-entity path no longer goes through `charge()`/`discharge()` and the unprotected burst route this PR guards against is gone.

One note in case you ever revisit it: `charge()` and `discharge()` return `power` even when the write is suppressed, so the manager subtracts it from the setpoint as if it had been applied. If a throttle comes back, that return should move inside the branch that actually sends.

Closing for now. Happy to reopen if the write path needs a rate limit again.

PR #1559, opened 2026-07-30
Add the SOCFULL bypass back to the discharge command

### LINKED ISSUE #1465: [Bug] Smart matching mode is not matching when solar power combined battery power is used
### Home Assistant Version

2026.06.2

newest master clone

### Describe the bug

Setup 3 Hypers each with solar panels and batteries. 
I noticed that the manager is not matching the P1 meter when there is not enough solar power available and battery energy is additionally needed (see log below, there is a always difference of at least 80W up to over 300W).
I assume there must be some miscalculation. 

### To Reproduce

smart matching mode, some solar power but not enough 

### Expected Behaviour

p1 meter matching to around 0W

Hyper2000

### Log Extracts

```shell

```

### Other Information

[home-assistant_zendure_ha_2026-07-01T15-52-14.250Z.log](https://github.com/user-attachments/files/29559303/home-assistant_zendure_ha_2026-07-01T15-52-14.250Z.log)

[home-assistant_zendure_ha_2026-07-01T16-55-13.902Z.log](https://github.com/user-attachments/files/29560960/home-assistant_zendure_ha_2026-07-01T16-55-13.902Z.log)

[home-assistant_zendure_ha_2026-07-01T17-43-47.681Z.log](https://github.com/user-attachments/files/29562202/home-assistant_zendure_ha_2026-07-01T17-43-47.681Z.log)

### LINKED ISSUE #1487: [Bug] The manager does not discharge the batteries if there is solar power generation
### Home Assistant Version

2026.7

V1.4.2

### Describe the bug

For some time now, the controller has not been charging my batteries when solar production is lower than my home’s energy consumption (see attached video).
It draws power from the external grid rather than using the batteries to make up for the shortfall in solar energy.

Image

Image

Image

Image

Image

### To Reproduce

on my installation ( one sf800 + 1 hyper 2000 and one sf1600ac + smart 3CT)
1 - have a solar production
2 - Start équipement who need some electricity
3 - look that he take electricity on external source

### Expected Behaviour

I'd like him to use my batteries instead of the power grid

Hyper2000

### Log Extracts

```shell

```

### Other Information

i can send video if you need

ahocevar:
Fixes #1465. Maybe also fixes #1487, which looks like the same defect reported from a different hardware mix.

A `SOCFULL` device passing solar through to the home has that pass-through booked as non-dispatchable and removed from the setpoint (the `discharge_bypass` accumulation in `powerChanged`, then `setpoint -= self.discharge_bypass`). That part is correct — the power reaches the house regardless of what the controller does.

`power_discharge` then distributes the remaining setpoint as **absolute** output targets. A device already exporting `B` watts of bypass is commanded its share of the *reduced* setpoint, i.e. `B` watts short of what it should deliver. The bypass is never added back, so the device is told to output roughly what it is already doing, nothing changes, and the grid covers the difference indefinitely.

At steady state every device sits at its command, so `sum(homeOutput) == setpoint`. Substituting into `setpoint = p1 + sum(homeOutput) - discharge_bypass` gives a falsifiable prediction:

> **p1 == discharge_bypass**

This is what the fix does in detail:

- Record the per-device bypass (`pwr_bypass`) alongside `discharge_bypass`.
- Express the distribution in dispatchable power: `dispatch_limit` / `dispatch_produced`, and per device `headroom = pwr_max - pwr_bypass`, `solar = -pwr_produced - pwr_bypass`.
- Add the bypass back when issuing the absolute command, and credit only the dispatchable part against the setpoint.
- `dev_start` subtracts only the dispatchable solar.

The setpoint arithmetic in `powerChanged` is untouched, so the #1151 guarantee ("with no device charging and `p1 >= 0` the setpoint stays `>= 0`") still holds structurally. Where no device is bypassing, every `pwr_bypass` is 0, all new terms collapse to the originals, and the distribution is unchanged.

No new constants, no per-device timing state, no signature changes.

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -137,6 +137,8 @@ class ZendureDevice(EntityDevice):
         self.maxSolar = 0
         self.pwr_max: int = 0
         self.pwr_produced: int = 0
+        # part of homeOutput the manager removes from its setpoint as non-dispatchable
+        self.pwr_bypass: int = 0
         self.actualKwh: float = 0.0
         self.state: DeviceState = DeviceState.OFFLINE
         self.exports_bypass: bool = True
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -427,6 +427,7 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
         onlinekWh = 0
 
         for d in self.devices:
+            d.pwr_bypass = 0
             if await d.power_get():
                 # get power production
                 d.pwr_produced = min(0, d.batteryOutput.asInt + d.homeInput.asInt - d.batteryInput.asInt - d.homeOutput.asInt)
@@ -451,7 +452,8 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
                     # skew), and subtracting more than was added fabricates a phantom negative
                     # setpoint — the root cause of #1151.
                     if d.state == DeviceState.SOCFULL and d.exports_bypass:
-                        self.discharge_bypass += min(-d.pwr_produced, home)
+                        d.pwr_bypass = min(-d.pwr_produced, home)
+                        self.discharge_bypass += d.pwr_bypass
                     self.discharge_limit += d.fuseGrp.discharge_limit(d)
                     self.discharge_optimal += d.discharge_optimal
                     self.discharge_produced -= d.pwr_produced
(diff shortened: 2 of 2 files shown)


ahocevar:
Worth noting this one is not sufficient on its own when the device is in bypass: it gets the right command but does not follow a raised `outputLimit`, so the demand still has to come from another device. #1565 covers that case.

CWILDMOUNTAIN:
well it looks like my devices are still not exiting bypass even with this fix (manager Calls for no Action) 
[home-assistant_zendure_ha_2026-08-07T15-08-32.252Z.log](https://github.com/user-attachments/files/30834538/home-assistant_zendure_ha_2026-08-07T15-08-32.252Z.log)

PR #1565, opened 2026-08-01
Do not expect dispatchable power from a device reporting bypass

ahocevar:
Builds on #1559 and has to merge after or instead of it — it uses the `pwr_bypass` and `dispatch_limit` introduced there, so until #1559 lands this PR shows its commit (5bdceda7606ff528d7f499e8a84bdf5be25d1a23) too. Also related to #1465.

A device reporting bypass does not follow a raised `outputLimit`. `power_discharge` nevertheless counts `pwr_max - pwr_bypass` as available on it, hands it the whole setpoint and looks no further, so the demand stays uncovered while other devices sit idle. `dev_start` cannot correct for it either: its threshold is `discharge_optimal * 2`, which with a single device discharging is 600 W, so any smaller deficit is never acted on at all.

`dispatchable()` now reads the `byPass` sensor — the same test `power_charge` already uses to avoid stopping bypassing devices — and returns 0 for a device in bypass, so both `dispatch_limit` and the per-device `headroom` reflect that nothing more can be had from it. `dev_start` then fires when the discharging devices are out of headroom as well as when they leave their optimal range.

The two go together: the `dev_start` term does nothing until `dispatch_limit` reflects the bypass. A device that does not report the property keeps `byPass` at 0, so `dispatchable()` returns what it did before and the distribution is unchanged for it.

### DIFF
--- a/custom_components/zendure_ha/device.py
+++ b/custom_components/zendure_ha/device.py
@@ -137,6 +137,8 @@ class ZendureDevice(EntityDevice):
         self.maxSolar = 0
         self.pwr_max: int = 0
         self.pwr_produced: int = 0
+        # part of homeOutput the manager removes from its setpoint as non-dispatchable
+        self.pwr_bypass: int = 0
         self.actualKwh: float = 0.0
         self.state: DeviceState = DeviceState.OFFLINE
         self.exports_bypass: bool = True
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -597,16 +597,22 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
 
         # Bypass already flows to the home and was subtracted from the setpoint, so it is
         # excluded from the capacities below and added back to the absolute command.
-        dispatch_limit = self.discharge_limit - self.discharge_bypass
+        # A device in bypass ignores an outputLimit above its pass-through, so it offers
+        # nothing dispatchable at all and another device has to cover the demand.
+        def dispatchable(d: ZendureDevice) -> int:
+            return 0 if d.byPass.asInt > 0 else max(0, d.pwr_max - d.pwr_bypass)
+
+        dispatch_limit = sum(dispatchable(d) for d in self.discharge)
         dispatch_produced = max(0, self.discharge_produced - self.discharge_bypass)
 
-        # distribute discharging devices, use produced power first, before adding another device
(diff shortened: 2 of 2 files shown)


ahocevar:
Running this on my own system since today, and the grid import in this situation is now covered. Since `dispatchable()` only engages when `byPass.asInt > 0`, that also confirms the Hyper 2000 does report the property.

CWILDMOUNTAIN:
I can confirm the issue, I will test the fix tomorrow and report back- thanks!

CWILDMOUNTAIN:
works pretty well- perfect

CWILDMOUNTAIN:
edit: well unfortunately I observed my devices are noch exiting bypass as the should 🤔
[home-assistant_zendure_ha_2026-08-07T15-08-32.252Z.log](https://github.com/user-attachments/files/30834573/home-assistant_zendure_ha_2026-08-07T15-08-32.252Z.log)

ahocevar:
@CWILDMOUNTAIN In my setup, there is another device in the mix which can supply the unmet demand (a 2400AC). This does not seem to be the case in your setup. You could try something like the following inserted at line 598 in `manager.py` on top of this pull request's branch:
```python
        # Bypass already flows to the home and was subtracted from the setpoint, so it is
        # excluded from the capacities below and added back to the absolute command.
        def capacity(d: ZendureDevice) -> int:
            return max(0, d.pwr_max - d.pwr_bypass)

        # A device in bypass ignores an outputLimit above its pass-through, so it offers
        # nothing dispatchable and another device has to cover the demand. That holds only
        # as long as another device can: bypass starts at socSet, which still leaves a full
        # battery to discharge from, so when every discharging device is in bypass and no
        # idle device is left to start, commanding them is the only option left - it may not
        # bring them out of bypass, but not commanding them certainly never will.
        no_alternative = (
            setpoint > 0
            and sum(capacity(d) for d in self.discharge if d.byPass.asInt == 0) == 0
            and not any(d.state != DeviceState.SOCEMPTY for d in self.idle)
        )

        def dispatchable(d: ZendureDevice) -> int:
            return capacity(d) if no_alternative or d.byPass.asInt == 0 else 0
```
Let me know if that command is obeyed by your Hypers. If so, it might make sense to add the above code to this pull request.

CWILDMOUNTAIN:
ah the code looks promising, I just implemented it and will test it and report back 👍

CWILDMOUNTAIN:
it's working like a charm- the devices are waking up as they should! 
thank you so much!

ahocevar:
Thanks for the feedback, @CWILDMOUNTAIN. I added the code section above to the pull request now.

PR #1567, opened 2026-08-02
feat(manager+tests): Manager modes tests. The multiple fixes in Manager (+min output feature)

igosoft:
CSV-driven spec tests for ZendureManager power distribution across all ManagerMode settings, single- and two-device scenarios.

  ## What

* More than 600 parameterized test cases spanning 6 manager modes. Each mode has a CSV spec (tests/manager_modes/data/<mode>.csv) that serves as the source of truth: every row is a case (case number as the first column) with the expected steady-state outputs per device (discharging / charging / device-to-grid). "any" SoC rows expand into three concrete cases (EMPTY / FULL / not full) so the collapse assumption is actually tested,   including cartesian expansion on the second device.
Two-device rows use optional columns: pv2_w…device2_to_grid_w for device 2, level/level2 for explicit electric levels, fuse_w for a shared fuse group, offgrid_w for off-grid consumers, exports=0 for the no-export-bypass configuration. Scenarios cover SoC-weighted splits, PV on one device only, negative P1 (export elsewhere), shared fuse groups, and off-grid devices.

* Minimum output power for Hub family devices (including tests).
* Device output compensation if one or more device isn't responding. 

  ## How it works

  - harness.py — FakeDevice with a physical battery plant model (solar --> battery --> grid), a residual P1 meter loop, and drive_metered() that runs real powerChanged() cycles to steady state for one or two devices.
  - Test files are thin: load CSV, run harness, assert every device against its spec. On failure a comparison table with the row's scenario note is printed (expected vs actual D/C/G per device).

  ## Modes covered

  | Mode | Rule |
  |---|---|
  | OFF | No commands issued, operation state = OFF |
  | MANUAL | Device outputs/draws exact manual power; P1 ignored |
  | MATCHING | Balances P to 0 with PV + battery |
  | MATCHING_DISCHARGE | max(self.produced, setpoint) -- discharge only, never charge |
  | MATCHING_CHARGE | min(produced, setpoint) -- solar pass-through only |
  | STORE_SOLAR | power_charge only, all solar goes to battery |

## Test Info
* **File:** `tests/manager_modes/harness.py`
* **Line:** 364
* **Error:** `AssertionError`

### Device Comparison Table

| device | PV | D exp | C exp | G exp | D act | C act | G act | Status |
| :--- | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :--- |
| **dev1** | 100 | 0 | 0 | 100 | 0 | 25 | 75 | <-- mismatch |
| **dev2** | 100 | 0 | 50 | 50 | 0 | 25 | 75 | <-- mismatch |


igosoft:
@Rolf-Smit @ahocevar I have implemented several fixes and added data-driven tests covering manager modes. These tests showed that your commits from your PR are also needed. I have grouped all the fixes into this single PR. Could you please review it?

Rolf-Smit:
@igosoft please point me to the relevant PR or commit that you used. I have multiple PRs open at the moment.

igosoft:
Indeed,  I used/adopted #1555 and #1559, #1565

igosoft:
@FireSon may you please also provide your feedback (on this PR) ?

ahocevar:
@igosoft Thanks for the heads-up. This pull request does not contain all of #1559 and #1565. I'm adding inline comments for details.

An alternative would be to get #1559 and #1565 merged first, and then you rebase this pull request on top of master. That would also surface the divergences I'm going to inline-comment on now.

custom_components/zendure_ha/manager.py
```diff
@@ -586,62 +632,109 @@ async def power_charge(self, setpoint: int, time: datetime) -> None:
...
                 self.pwr_low = 0 if (delta := d.discharge_start * 1.5 - pwr) <= 0 else self.pwr_low + int(delta)
                 pwr = 0 if self.pwr_low > d.discharge_optimal else pwr
 
-            setpoint -= await d.power_discharge(pwr)
-            dev_start += 1 if pwr != 0 and d.electricLevel.asInt + 3 < self.idle_lvlmax else 0
+            # Credit the ACTUAL dispatched power, not the pre-floor pwr computed above:
+            # a device's own min_output floor (device.py) can raise its output past
+            # what was commanded here, and the remaining devices must not be handed
+            # the setpoint a floored peer already covers.
+            actual = await d.power_discharge(pwr + d.pwr_bypass) - d.pwr_bypass
```
ahocevar: The `max(0, ...)` got lost here:


```suggestion
            actual = max(0, await d.power_discharge(pwr + d.pwr_bypass) - d.pwr_bypass)
```

custom_components/zendure_ha/manager.py
```diff
@@ -586,62 +632,109 @@ async def power_charge(self, setpoint: int, time: datetime) -> None:
...
-            if len(self.discharge) > 1 and i == 0 and d.state != DeviceState.SOCFULL:
+                pwr = max(setpoint - limit, 0 if d.state != DeviceState.SOCFULL else solar)
+            pwr = min(pwr, setpoint, headroom)
+
+            # make sure we have devices in optimal working range; only park the lowest-SOC
+            # device when it is not producing and the other devices can actually absorb the
+            # setpoint. Bypassing peers contribute nothing dispatchable, so zeroing here would
+            # leave the demand uncovered and the device is restarted from idle on the next
+            # cycle - an endless start/stop cycle.
+            if i == 0 and d.state != DeviceState.SOCFULL and d.pwr_produced == 0 and dispatch_limit - headroom >= setpoint:
```
ahocevar, line 829: The `dispatch_limit - headroom >= setpoint` condition is now only reachable for non-producing devices, because of the added `d.pwr_produced == 0` condition. That's narrower than what I had in mind, but probably ok.

igosoft:
Thanks for the comments, @ahocevar. It doesn't, but it is based on the code from those PRs. No problem, I can also rebase it if #1559 and #1565 get merged

custom_components/zendure_ha/manager.py
```diff
@@ -586,62 +632,109 @@ async def power_charge(self, setpoint: int, time: datetime) -> None:
...
+                if self.discharge_weight != 0:
+                    share = int(battery_budget * device_weight / self.discharge_weight)
+                elif len(self.discharge) > i:
+                    share = int(battery_budget / (len(self.discharge) - i))
+                else:
+                    share = 0
+                self.discharge_weight -= device_weight
+
+                # own solar (solar-first floor, #1555) + SoC-weighted battery-budget share.
+                pwr = solar + share
```
ahocevar, line 816: This is broader than my `if pwr < solar and d.state == SOCFULL: pwr = solar`, but I guess that's ok.

tests/manager_modes/harness.py
```diff
@@ -0,0 +1,439 @@
...
+
+        self.solarInput = _sensor(pv)
+        self.homeOutput = _sensor(0)
+        self.homeInput = _sensor(0)
+        self.batteryOutput = _sensor(0)   # packInputPower: battery -> out (discharge)
+        self.batteryInput = _sensor(0)    # outputPackPower: into battery (charge)
+        self.electricLevel = _sensor(level)
+        self.minSoc = _sensor(0)
+        # a full battery passing its solar reports hardware bypass, like real devices
+        self.byPass = _sensor(1 if (soc_state == DeviceState.SOCFULL and pv > 0) else 0)
```
ahocevar: The case #1565 exists for (a not-full device reporting bypass and ignoring a raised outputLimit) never occurs in any CSV row, and there is no case where all discharging devices are in bypass with no idle device left, so `no_alternative` is never exercised. A regression to `dispatchable()` for non-full bypass would pass the suite.

custom_components/zendure_ha/manager.py
igosoft: Indeed, some test cases still failed (I don't remember now), thanks for reviewing the change

custom_components/zendure_ha/manager.py
igosoft: ok

tests/manager_modes/harness.py
igosoft: good points, I added cases to cover it

Issue #1594, opened 2026-08-13
[Bug] Bug report – SF2400 AC Smart Charge/Discharge and local IP communication


sunnybeam1969:
### Home Assistant Version

2026.5.2

### Zendure Integration Version

v1.4.4

### Describe the bug

Title: Wrong local IP address used for SolarFlow 2400 AC

I am using a Zendure SolarFlow 2400 AC with the HACS Zendure Home Assistant Integration v1.4.4.

The integration repeatedly generated the following error:

Logger: custom_components.zendure_ha.device
Source: custom_components/zendure_ha/device.py:831

ClientConnectorError for SolarFlow 2400 AC during httpPost:
Cannot connect to host 192.168.68.80:80
[Connect call failed ('192.168.68.80', 80)]

The actual IP address of my Zendure AC2400 is:

192.168.68.132

My MQTT server is:

192.168.68.112

In device.py, the original code is:

self.ipAddress = definition.get("ip", "") if definition.get("ip", "") != "" else f"zendure-{definition['productModel'].replace(' ', '')}-{self.snNumber}.local"

It appears that definition["ip"] contained the incorrect address 192.168.68.80.

As a test, I temporarily changed the line to:

self.ipAddress = "192.168.68.132"

After doing this, the problem was immediately resolved.

Smart Charging now works correctly and the AC2400 actually charges. For example, the charging power is currently around 593 W.

The following sensor also reports the actual charging power correctly:

sensor.solarflow_2400_ac_output_pack_power

Conclusion:

The problem appears to be related to the wrong IP address being provided in definition["ip"], rather than the Smart Charging logic or MQTT connection.

It would be useful to investigate where definition["ip"] is populated and why it contains 192.168.68.80 when the actual device IP is 192.168.68.132.

For now, I have hardcoded the correct IP address as a workaround, but obviously a permanent solution would be preferable so that this does not happen again after a HACS update.

Thanks for looking into this!

### To Reproduce

I am not sure yet how to reproduce the incorrect IP address being assigned.

The issue appeared after an update of the Zendure integration. My SolarFlow 2400 AC had the correct IP address on my network (192.168.68.132), but the integration was trying to connect to 192.168.68.80 on port 80.

The error occurred repeatedly when Home Assistant tried to send commands to the SolarFlow 2400 AC, especially while using Smart Charging.

I did not manually configure 192.168.68.80 anywhere in Home Assistant or in the Zendure integration.

The important part is that the original code uses:

[...]

Issue #1583, opened 2026-08-10
zenSDK connects to a stale IP from the cloud deviceList, with no way to override it


jarivanb:
### Summary

My SolarFlow 2400 AC got a new DHCP lease. The integration kept trying the old address and there is no setting anywhere to correct it.

- Integration: 1.4.3 | Home Assistant: 2026.8.1
- Device: SolarFlow 2400 AC, connection mode zenSDK
- Old IP 192.168.0.43, actual IP 192.168.0.19

### What happens

```
ClientConnectorError for SolarFlow 2400 AC during httpPost:
Cannot connect to host 192.168.0.43:80 ssl:default [Connect call failed]
```

Hundreds of these, on both httpPost and httpGet. The Zendure app shows the correct IP-address for the same device.

All writes fail silently: `ac_mode` stuck on `input`, output limit stuck at 0 W, no discharge all day at 99 percent SOC. Telemetry keeps arriving over cloud MQTT, so nothing looks wrong on the dashboard.

Reloading, removing and re-adding the integration, restarting HA, reconnecting the battery to WiFi and a full power cycle all made no difference.

### Cause

`device.py` line 120 takes the address from the `ip` field of `POST /api/ha/deviceList`. That cloud record was stale for hours. The mDNS hostname fallback exists but only applies when the cloud returns an empty string, so it is never reached.

The IP is therefore dictated by a cloud value users cannot see or change.

Mrdauphi:
I'm struggling with exact this issue after my 2 800 plus batteries got a new IP from the DHCP.  At the moment only the cloud connection seem to work for me.   Did you find a workaround to get the zenSDK connection working again?

jarivanb:
Hi, in the meantime I deleted my battery from the app and my account and added it again. That seems to trigger the IP-reset on their servers.

PR #1596, opened 2026-08-14
Improve ZenSDK connectivity with configurable per-device IPs when mDNS or cloud IPs are unreliable

Strudeldew:
Cloud deviceList metadata can contain stale private IP addresses after router or subnet changes. In that case, local ZenSDK polling and writes time out even though the device is reachable via its mDNS hostname.

This change makes local HTTP host selection resilient by trying mDNS first and using the cloud IP as a fallback.

Refs: #1583, #1594, #1615

### DIFF
--- a/custom_components/zendure_ha/config_flow.py
+++ b/custom_components/zendure_ha/config_flow.py
@@ -8,6 +8,7 @@ from typing import Any
 import voluptuous as vol
 from homeassistant.config_entries import ConfigFlow, ConfigFlowResult, OptionsFlow
 from homeassistant.core import callback
+from homeassistant.data_entry_flow import section
 from homeassistant.exceptions import HomeAssistantError
 from homeassistant.helpers import selector
 
@@ -15,14 +16,18 @@ from .api import Api, ApiError
 from .const import (
     CONF_APPTOKEN,
     CONF_AUTO_MQTT_USER,
+    CONF_LOCAL_IP_SECTION,
+    CONF_LOCAL_IPS,
     CONF_MQTTLOCAL,
     CONF_MQTTLOG,
     CONF_MQTTPORT,
     CONF_MQTTPSW,
     CONF_MQTTSERVER,
     CONF_MQTTUSER,
+    CONF_OVERWRITE_IP,
     CONF_P1METER,
     CONF_SIM,
+    CONF_USE_MDNS,
     CONF_WIFIPSW,
     CONF_WIFISSID,
     DOMAIN,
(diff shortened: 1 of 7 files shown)


jarivanb:
This would make so much sense to have this approach! Thanks!

jcjveraa:
I'm actually looking in to this topic as my mDNS seems to be unstable, I get a bunch of these in my logs every day:


`ClientOSError for 2400-1 during httpPost: [Errno 104] Connection reset by peer
TimeoutError for 2400-2 during httpPost!
TimeoutError for 2400-1 during httpPost!
ClientConnectorDNSError for 2400-1 during httpPost: Cannot connect to host zendure-solarflow2400ac+-hec4nencn520756.local:80 ssl:default [MDNS lookup failed, Misformatted domain name]
ClientConnectorDNSError for 2400-2 during httpPost: Cannot connect to host zendure-solarflow2400ac+-hec4nencn520417.local:80 ssl:default [MDNS lookup failed, Misformatted domain name]
`

As I think many hass users do, I've assigned static IP addresses to both units. 

mDNS is nice but a little error prone in many network setups - I'd much appreciate an easy option (outside of the manual configuration overrides you can do) to specify the ip address of the units one would want to connect. 

If this is merged I'd expect even more errors, so please include an "opt out" thru the config flow options where the users can then manually specify a local ip (or confirm the local ip that was discovered via mDNS to have best of both worlds). Include a warning that this is an advanced user option as it means the ips need to be static.

Strudeldew:
That sounds reasonable. I’ve never had any issues with mDNS myself, so I hadn’t considered that it could be problematic in some network setups. I’ll have a look at this and see how we can best handle a manually specified IP address, then update the PR.

jcjveraa:
Thanks for taking the time to include that :) 

I hope to be able to find the time to pull your branch in my lunch break this Monday to see if it works for me. 

In the meantime I discovered that my error messages were somewhat of a red herring as per #1635 which I raised. Turns out this integration doesn't do a proper escape on the + character in the product name for making the mDNS string, ~~but next to that Zendure did some trickery where the mDNS name of the device includes an escaped # (which is \043) instead of an escaped + (\053), as I show in that issue. That "off by 10" in the escaped value makes me wonder if this was intentional or sloppy.~~ **Gemini tricked me, ignore that bit about the wrong code. 43 is the correct decimal encoding of a plus sign, had not expected the AI to hallucinate that. Need to fix this in the issue too...**

Likely it would affect any model with a + or something similar in the product name.

PR #1613, opened 2026-08-21
fix(manager): don't let devices at a SoC limit mask a real grid deficit

christophecollet78:
In MATCHING mode with a mixed-SoC fleet, `dev_start` in power_discharge() credits every device in `self.discharge` with `discharge_optimal` of headroom, and subtracts the whole of `discharge_produced`. Two of those inputs are not dispatchable:

- A SOCFULL device cannot supply anything from its battery, and its solar bypass was already netted out of the setpoint in powerChanged() — subtracting it again in `discharge_produced` double-counts it.
- A SOCEMPTY device is pinned at minSoc. It still lands in `self.discharge` because it passes its own solar through (homeOutput > 0), but it has no headroom either.

Either one cancels the real deficit, `dev_start` stays 0, no idle device is started, and the grid covers the gap indefinitely. Devices charging from their own DC solar are filed under `idle` (homeInput == homeOutput == 0), so the "stop charging devices" loop never touches them — starting an idle device is the only lever, and it was blocked.

Separately, `pwr = min(pwr, setpoint, d.pwr_max)` undid the SOCFULL floor set a few lines above, commanding a bypassing device below what it already produces. That only curtails its PV and grows the import, then rebounds next cycle — visible as the manager state flipping between charge/discharge/idle dozens of times per hour.

Measured on a 3-device fleet (2x Hyper 2000 + SolarFlow 800 Plus, local MQTT, Shelly Pro 3EM as P1):

  2026-08-19 15:30 - one Hyper at 100% bypassing 590 W, the two others charging 543 W
  from their own solar, house 894 W, total PV 1172 W, 304 W bought from the grid. PV
  exceeded the load by 278 W throughout.

  2026-08-20 09:06 - both Hypers at minSoc passing 211 W of solar through, SolarFlow at
  7% with reserve left but charging, 159 W bought. After the fix, same solar and same
  load 11 minutes later: SolarFlow delivering 136 W, grid at -11 W.

Three scenario tests added to tests/test_manager_dispatch.py, using the existing harness. Each one fails on the code before this change and passes after it. I ran the file with the pinned test stack (pytest-homeassistant-custom-component==0.13.347, pytest==9.0.3, pytest-asyncio==1.4.0) against homeassistant==2026.7.3: 6 passed — the three new tests plus the three pre-existing ones in that file, unchanged. Note that this pin of pytest-homeassistant-custom-component requires Python 3.14, it does not resolve on 3.13.

Not changed here: SOCFULL devices still take a weighted share of the dispatchable setpoint via `discharge_weight`, so the devices that can actually discharge converge slowly. And power_charge() has the mirror-image issue with `charge_optimal` counting SOCFULL devices. Both looked like design calls rather than bugs, so they are left alone.

### DIFF
--- a/custom_components/zendure_ha/manager.py
+++ b/custom_components/zendure_ha/manager.py
@@ -600,7 +600,13 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
             await d.power_discharge(0 if max(0, d.pwr_offgrid) == 0 else 10)
 
         # distribute discharging devices, use produced power first, before adding another device
-        dev_start = max(0, setpoint - self.discharge_optimal * 2 - self.discharge_produced) if setpoint > SmartMode.POWER_START else 0
+        # Only *dispatchable* capacity may offset the setpoint. A device at either SoC limit has none:
+        # SOCFULL cannot supply from its battery (its bypass was already netted out in powerChanged),
+        # SOCEMPTY is pinned at minSoc. Crediting them cancels a real deficit and the grid pays instead.
+        blocked = (DeviceState.SOCFULL, DeviceState.SOCEMPTY)
+        blocked_optimal = sum(d.discharge_optimal for d in self.discharge if d.state in blocked)
+        dispatchable = (self.discharge_optimal - blocked_optimal) * 2 + (self.discharge_produced - self.discharge_bypass)
+        dev_start = max(0, setpoint - dispatchable) if setpoint > SmartMode.POWER_START else 0
         solaronly = self.discharge_produced >= setpoint
         limit = self.discharge_produced if solaronly else self.discharge_limit
         setpoint = min(limit, setpoint)
@@ -628,6 +634,11 @@ class ZendureManager(DataUpdateCoordinator[None], EntityDevice):
             if limit < setpoint - pwr:
                 pwr = max(setpoint - limit, 0 if d.state != DeviceState.SOCFULL else -d.pwr_produced)
             pwr = min(pwr, setpoint, d.pwr_max)
+
+            # Re-apply the SOCFULL floor: the min() above clamps to a setpoint that already had this
+            # device's bypass removed. Commanding it below its own production only curtails the PV.
+            if d.state == DeviceState.SOCFULL:
+                pwr = max(pwr, min(-d.pwr_produced, d.pwr_max))
 
             # make sure we have devices in optimal working range
             if len(self.discharge) > 1 and i == 0 and d.state != DeviceState.SOCFULL:
--- a/tests/test_manager_dispatch.py
+++ b/tests/test_manager_dispatch.py
@@ -62,6 +62,7 @@ def _make_manager_harness() -> ZendureManager:
     manager.charge_last = datetime.min
     manager.charge_weight = 0
     manager.discharge = []
+    manager.discharge_bypass = 0
     manager.discharge_produced = 0
     manager.discharge_limit = 0
     manager.discharge_optimal = 0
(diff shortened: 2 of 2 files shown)


igosoft:
Hi Christopher 

One of these bugs have been already fixed on https://github.com/Zendure/Zendure-HA/pull/1567  
Could you please have a look ?

However, the one blocking idle-start via SOCFULL/SOCEMPTY masking a real grid deficit is not. Feel free to commit the necessary changes and tests so we can wrap everything up in a single PR covering Manager modes

custom_components/zendure_ha/manager.py
```diff
@@ -600,7 +600,13 @@ async def power_discharge(self, setpoint: int) -> None:
             await d.power_discharge(0 if max(0, d.pwr_offgrid) == 0 else 10)
 
         # distribute discharging devices, use produced power first, before adding another device
-        dev_start = max(0, setpoint - self.discharge_optimal * 2 - self.discharge_produced) if setpoint > SmartMode.POWER_START else 0
+        # Only *dispatchable* capacity may offset the setpoint. A device at either SoC limit has none:
+        # SOCFULL cannot supply from its battery (its bypass was already netted out in powerChanged),
+        # SOCEMPTY is pinned at minSoc. Crediting them cancels a real deficit and the grid pays instead.
+        blocked = (DeviceState.SOCFULL, DeviceState.SOCEMPTY)
```
igosoft: Could you please consider moving blocked variable to const.py and naming it BLOCKING_DEVICE_STATES ?

christophecollet78: Done, thanks — moved to `BLOCKING_DEVICE_STATES` in `const.py`, right after the
`DeviceState` enum, with a short comment on why those two states are grouped.

`DeviceState` is still imported in `manager.py` for the SOCFULL floor a few lines
further down, so the import stays.

igosoft (approved):

tkorner:
I reviewed this from the code only (no Zendure hardware here), plus a local run of the test suite. Summary: the diagnosis is right and the second hunk is clearly correct, but I think `blocked_optimal` is doing one thing more than it should, and there is a corner where the same masking survives the fix. Details and repro steps below.

## Reproduction

Checked out `d15e9f0`, installed the pinned stack from `requirements_test.txt` on Python 3.14.7:

- on the PR as-is: **17 passed**
- reverting only `manager.py` and `const.py` to `master` (`e71b0b8`) while keeping the new tests: **3 failed, 3 passed** — the three new tests fail with exactly the described symptoms (`assert 304 >= 590`, `assert ([] or [])`, `assert []`), the three pre-existing ones are untouched.

So "fails before, passes after" holds independently. Thanks for writing the scenarios against real captures — it made this reviewable without hardware.

(All line numbers below refer to the base commit `e71b0b8`, not to the patched file.)

## The arithmetic is safe where it has to be

To answer the obvious question about `max(0, ...)`: `dispatchable` cannot go negative.

- `self.discharge_optimal` is accumulated over exactly the members of `self.discharge` (manager.py:462), and `d.discharge_optimal` is `discharge_limit // 4 >= 0`. So `self.discharge_optimal - blocked_optimal` is a sum over a subset of non-negative terms — it reaches exactly `0` when every discharging device is blocked, never below.
- `discharge_bypass` adds `min(-d.pwr_produced, home)` (manager.py:460) while `discharge_produced` adds the full `-d.pwr_produced` (manager.py:463) over a superset of devices, and `pwr_produced = min(0, ...) <= 0`. So `discharge_produced - discharge_bypass >= 0` term by term.

With every device blocked, `dev_start` lands at `setpoint - (discharge_produced - discharge_bypass)`, i.e. the deficit minus pass-through solar, which is the intended reading. The `max(0, ...)` is never load-bearing on the low side.

It is worth writing down why the `- self.discharge_bypass` is needed, because it is the part I find most convincing. Per device, the contribution to `dev_start` reduces to `homeOutput - (-pwr_produced)`: the `+home` added to the setpoint in `powerChanged` (manager.py:465), the `-bypass` removed at manager.py:488, and the `+bypass` added back here cancel exactly. A device that exactly passes its own solar through nets out to zero. That is the correct model, and `master` did not have it.

## 1. The #1151 cap leaves a residual that is still credited as capacity

The cancellation above is exact only while `-d.pwr_produced == homeOutput`. When production exceeds the homeOutput actually contributed — the internal-trickle-charge / sensor-skew case the cap at manager.py:460 exists for — the per-device term becomes `home - (-pwr_produced) < 0`. That is negative demand: phantom capacity equal to the residual, on the capacity side of a subtraction it was never on the demand side of.

For a SOCFULL bypassing device at 600 W solar of which 400 W reaches the home bus: `discharge_bypass = min(600, 400) = 400`, `discharge_produced = 600`, so 200 W stays in `discharge_produced - discharge_bypass` and is credited as dispatchable. With 150 W genuinely bought from the grid:

```
setpoint     = 150 + 400 - 400 = 150
dispatchable = (300 - 300) * 2 + (600 - 400) = 200
dev_start    = max(0, 150 - 200) = 0
```

No idle device is started and the 150 W stays on the grid — the same failure mode this PR fixes, one layer down. Repro against your own harness:

```python
def _scenario(p1: int):
    """SOCFULL device: 600 W solar, but only 400 W reaches the home bus."""
    manager = _make_manager_harness()

    full = _FakeDevice(pwr_max=1200, electric_level=100, state=DeviceState.SOCFULL)
    full.pwr_produced = -600   # full solar input, per manager.py:439-440
    full.discharge_start = 120
    full.discharge_optimal = 300
    home = 400                 # homeOutput actually delivered

    reserve = _FakeDevice(pwr_max=800, electric_level=60)
    reserve.discharge_start = 80
    reserve.discharge_optimal = 200

    manager.discharge = [full]
    manager.discharge_limit = 1200
    manager.discharge_optimal = full.discharge_optimal
    manager.discharge_produced = -full.pwr_produced           # 600, manager.py:463
    manager.discharge_bypass = min(-full.pwr_produced, home)  # 400, manager.py:460 cap
    manager.discharge_weight = full.pwr_max * full.electricLevel.asInt
    manager.idle = [reserve]
    manager.idle_lvlmax = manager.idle_lvlmin = 60
    return manager, reserve, p1 + home - manager.discharge_bypass


async def test_residual_swallows_a_real_deficit() -> None:
    manager, reserve, setpoint = _scenario(p1=150)
    await manager.power_discharge(setpoint)
    assert reserve.discharge_calls, (
        "no idle device started: the 200 W residual (600 W produced - 400 W capped "
        "bypass) is credited as dispatchable and cancels a real 150 W deficit"
    )
```

Fails on this branch. Both new SOCFULL scenarios use `discharge_produced == discharge_bypass == 590`, so the capped case is not covered.

Direction of a fix, without prescribing one: the solar term wants to be built the same way `powerChanged` built the setpoint — per device capped at its `homeOutput` contribution — so the residual cannot re-enter as capacity. Doing that inside `power_discharge` rather than changing `discharge_produced` itself would keep `solaronly` and `limit` untouched.

## 2. `SOCFULL` in `BLOCKING_DEVICE_STATES` contradicts the rest of `power_discharge`

The `SOCEMPTY` half is airtight, and the code already agrees with it: the idle-start loop at manager.py:644 explicitly skips `SOCEMPTY`. This PR makes `dev_start` consistent with a rule the function already enforces.

`SOCFULL` is the opposite case — the same function treats it as dispatchable in three places:

- manager.py:644 starts an **idle** `SOCFULL` device to serve a deficit; only `SOCEMPTY` is excluded. After this change, a discharging `SOCFULL` device is credited with zero headroom for the very deficit that an idle `SOCFULL` device is then started to cover.
- the weight at manager.py:614 is `pwr_max * SoC`, so a `SOCFULL` device draws the **largest** share of the setpoint, and manager.py:622 is a floor, not a ceiling — nothing caps it at its own production. Your PR description acknowledges this ("still take a weighted share").
- `DeviceState.SOCFULL` is `socLimit == SOCFULL or electricLevel >= socSet` (device.py:646) — "at the user's configured max SoC", which is not the same as "in bypass". The bypass accounting at manager.py:459 correctly qualifies it as `state == SOCFULL and d.exports_bypass`; `BLOCKING_DEVICE_STATES` drops that qualifier. A device that is full but not exporting bypass has a full battery and plenty of headroom, and gets zeroed anyway.

I measured how much this half carries. Narrowing the constant to `(DeviceState.SOCEMPTY,)`:

```
5 passed, 1 failed
FAILED test_socfull_bypass_does_not_mask_a_real_deficit_from_idle_devices
```

So the de-double-counting and the `SOCEMPTY` fix and the floor re-application all stand on their own; only that one test depends on treating `SOCFULL` as having no battery headroom. If the premise is "a device in hardware bypass will not discharge its battery even when commanded", then keying on `d.exports_bypass` / `d.byPass.asInt` rather than on the SoC state would express it precisely and would leave full-but-not-bypassing devices alone. If instead the premise is genuinely about SoC, then manager.py:644 looks like it should skip `SOCFULL` too — right now the two rules disagree.

Is there a capture behind the `SOCFULL` half specifically, i.e. a device at `socSet` that was commanded above its production and did not deliver?

## 3. Minor: harness fidelity

`_FakeDevice.power_discharge` returns `power` verbatim, whereas the real `ZendureDevice.power_discharge` does `max(0, min(power, discharge_limit))` and returns `homeOutput` on the no-action path (device.py:672-678). Two consequences for reading the new tests:

- the new floor lets a `SOCFULL` device's command exceed `setpoint`, so `setpoint -= ...` at manager.py:637 can go negative for devices sorted after it. That is harmless against the real device, which clamps to `0` — but the harness cannot show it either way. Note the sort is by `electricLevel` ascending, and `socSet` is per device, so a `SOCFULL` device at 85 % can precede a non-`SOCFULL` device at 95 %; it is not guaranteed to be last.
- `setpoint` bookkeeping across the loop is idealised relative to the device's actual return value.

Both are pre-existing properties of the harness, not something this PR introduces.

## Merge order

Five open PRs touch `manager.py`. I test-merged each against this branch and ran the suite.

| PR | merge with #1613 | note |
|----|------------------|------|
| #1596 | clean | only imports and `loadDevices` (manager.py:27-135). Fully independent. |
| #1555 | clean, **but semantically contradictory** | see below |
| #1488 | clean, **but breaks the build** | see below |
| #1567 | conflict | exactly the two hunks of this PR |

**#1555** replaces `setpoint -= self.discharge_bypass` with a separate `dispatchable` used only on the charge path, and deliberately passes the **full** setpoint to `power_discharge` ("outputLimit is an absolute output target and not a delta on top of the bypass"). That removes the premise this PR's `- self.discharge_bypass` rests on. Merged together, git is happy and `17 passed`, but the bypass is then counted against the fleet twice — once by staying in the demand, once by being removed from the capacity — and `dev_start` is overstated by exactly `discharge_bypass`. With the 590 W capture in this PR, that is a device started on every cycle a bypassing device is present. The tests do not catch it because they call `power_discharge` directly and never go through `powerChanged`. These two need to be reconciled by hand whichever lands first. (There would also be two different `dispatchable` locals in one file.)

**#1488** changes the signature to `power_discharge(self, setpoint, time)` without updating `tests/test_manager_dispatch.py`. On `master` alone it is already **2 failed, 12 passed**; merged with this PR, **5 failed, 12 passed** — all `TypeError: missing 1 required positional argument: 'time'`. It needs a rebase regardless of what happens here.

**#1567** is the interesting one: it independently derives the same root cause — same `discharge_produced - discharge_bypass` de-double-counting, same re-applied `SOCFULL` floor — but blocks on the actual bypass state (`byPass.asInt == 0`, `capacity(d) = max(0, d.pwr_max - d.pwr_bypass)`) rather than on `DeviceState`, which is the distinction in point 2 above. Two contributors converging on this independently is a good sign the bug is real.

It does **not** subsume this PR, though. I ported the `SOCEMPTY` scenario from this branch onto #1567's own harness and it fails there: `capacity()` credits a `SOCEMPTY` device with its full `pwr_max`, so `dev_start = max(0, 370 - min(2400, 1200 + 211)) = 0` and no idle device is started. The `SOCEMPTY` masking survives #1567 despite its 630-test suite.

Suggested order, for what it is worth: **#1613 first** — it is small, isolated, green, and independent of #1596/#1555 textually. Then #1488 after a rebase that fixes the signature and the call sites. #1567 has to rebase onto whatever lands anyway, and should carry the `SOCEMPTY` case forward into its `manager_modes` suite. #1555 and #1613 need an explicit decision about whether `power_discharge` receives the setpoint with or without the bypass — that is one call by the maintainer, and it unblocks both.

PR #1650, opened 2026-09-09
Stop calling the deprecated device_registry.async_get_device

DerkSchooltink:
## Problem

`device_registry.async_get_device` is deprecated — identifiers are no longer unique across config entries — and is **removed in HA 2027.8.0**. Until then it logs a warning on every call:

```
Detected that custom integration 'zendure_ha' calls `device_registry.async_get_device`,
which is deprecated ...; use `async_get_device_by_identifier`, ... instead
```

We call it in four places, all single-identifier lookups of the shape `{(DOMAIN, x)}`: `EntityDevice.__init__` and `EntityDevice.updateVersion` in `entity.py`, and twice in `Migration.check_device` (the identifier-candidate loop).

## Fix

Route all four through one helper, `entity.device_by_identifier`.

The replacement, `async_get_device_by_identifier(identifier, config_entry_id)`, only exists from HA 2026.8.0, and `hacs.json` declares a minimum of 2025.4.0 — so swapping the calls outright would break every install on an older core. The helper prefers the new API and falls back where it is absent:

```python
registry = dr.async_get(hass)
if (entries := hass.config_entries.async_entries(DOMAIN)) and hasattr(registry, "async_get_device_by_identifier"):
    return registry.async_get_device_by_identifier((DOMAIN, identifier), entries[0].entry_id)
return registry.async_get_device(identifiers={(DOMAIN, identifier)})
```

Scoping to our entry returns the same device: identifiers are unique *within* an entry, and `manifest.json` sets `single_config_entry`, so there is exactly one to scope to. I did not thread `config_entry_id` through the constructors instead — that would touch ~20 `super().__init__` chains across 12 device files for a warning fix.

## Testing

The pinned test env is HA 2026.7.3, which predates the replacement, so CI alone only exercises the fallback branch. Ran the suite both ways: **18 passed on 2026.7.3** (fallback) and **18 passed on 2026.9.0** (entry-scoped).

On 2026.9.0, with `report_usage` instrumented, both lookups returning the same device:

```
old async_get_device  -> found=True  deprecation_reports=1
device_by_identifier  -> found=True  deprecation_reports=0
```

`tests/test_device_registry_lookup.py` covers both branches — the one CI's pin cannot reach uses a fake registry, so neither side bitrots silently.

## Notes

I used a feature probe (`hasattr`) rather than a version check — no version parsing, but it costs the return type and means one branch is always fake-tested. Happy to switch to `AwesomeVersion(HA_VERSION) >= "2026.8.0"`, or to raise `hacs.json` to 2026.8.0 and call the new API directly (~4 lines, but drops 2025.4–2026.7 users), if you prefer either.

## Not covered

`DeviceInfo["via_device"]` in `EntityDevice.__init__` is also deprecated (removed in 2027.8.0, replaced by `via_device_id`). That one needs the parent's *registry id* rather than an identifier tuple, and the parent is not guaranteed to be registered when the child is constructed — it wants its own change.


davinkevin:
Confirming the `via_device` half on **HA 2026.9.1** (custom integration loads fine, warnings only).

Five deprecation warnings fire on startup, and they split across the two changes described here:

Covered by this issue — `device_registry.async_get_device`:
- `migration.py:35` — `existing = device_registry.async_get_device(identifiers={(DOMAIN, identifier)})`
- `entity.py:229` — `if di := device_registry.async_get_device(identifiers={(DOMAIN, sn)}):`
- `entity.py:357` — `device_entry = device_registry.async_get_device(identifiers={(DOMAIN, identifier)})`

Not covered (the "Not covered" section above) — `device_registry.async_get_or_create` with a deprecated `via_device` parameter:
- `sensor.py:53` — `self.add([self])`
- `binary_sensor.py:37` — `self.add([self])`

Both are removed in 2027.8.0. Would it help to track the `via_device` → `via_device_id` migration as its own issue, or do you prefer to fold it in here?

DerkSchooltink:
Thanks for confirming on 2026.9.1, and for the per-site breakdown — it matches: the three `async_get_device` hits are this PR (the fourth call site, the fallback loop in `check_device`, only fires when the first lookup misses), and the two at `sensor.py:53` / `binary_sensor.py:37` are `self.add([self])` handing `EntityDevice.__init__`'s `via_device` to the core.

Rather than an issue, I opened the `via_device` → `via_device_id` half as its own PR: #1651, stacked on this branch. Kept it separate because it is not a rename: the id has to be resolved at construction instead of by the core at add time, and an unresolvable `via_device_id` raises and fails the entity add where `via_device` only warned, so it needs a guard for the unregistered-parent case plus its own version gate (`via_device_id` is also 2026.8+). This PR stays a pure lookup swap with no behaviour change.

PR #1663, opened 2026-09-20
Bound the charge hold-off so repeated flips cannot lock out charging

igotsoul:
## Problem

`power_charge` holds charging off for 60 s after a discharge→charge flip ("prevent hysteria"). `power_discharge` resets `charge_time` to `datetime.max`, and the next `power_charge` re-arms a **fresh** 60 s window. The window restarts on every counter-event and has no upper bound.

If the grid point keeps flipping faster than 60 s, the window never expires and the battery does not charge at all for as long as the excitation lasts, while surplus is exported.

This was analysed in detail with logs and 30 days of recorder data in #1525 (closed by the stale bot, no fix). Typical excitation sources reported there and in #1461:

- thermostat-cycling appliances (iron: off-phases avg. 53 s, 28 of 35 saw no charge command)
- cloud flicker on the PV side
- PV-surplus EV charging (e.g. evcc), where the wallbox keeps re-adjusting its current and the grid point crosses zero repeatedly

## Change

The hold-off is still re-armed after every discharge, but it is now bounded by the **first arming of the episode**:

- `charge_hold_start` records when the current flip episode started
- `charge_time = min(now + delay, charge_hold_start + CHARGE_HOLDOFF)`
- the episode ends with the first released charge, so the next flip starts a new episode with a full hold-off

Net effect: at most 60 s without charging per episode, instead of unbounded. Anti-flapping intent and the 60 s / 2 s / 300 s values are unchanged; they are moved into `SmartMode` constants (`CHARGE_HOLDOFF`, `CHARGE_HOLDOFF_SHORT`, `CHARGE_HOLDOFF_RESET`).

One behavioural detail: if a re-arm happens after the bound has already passed, charging is released on that same call instead of one tick later.

## Tests

Added to `tests/test_manager_dispatch.py`:

- `test_power_charge_single_flip_keeps_full_holdoff`: single flip still gets the full 60 s (unchanged behaviour)
- `test_power_charge_holdoff_is_bounded_under_repeated_flips`: regression for #1525; flips at +35 s and +55 s do not extend the window beyond +60 s, a flip at +70 s releases immediately. **Fails on current master.**
- `test_power_charge_new_episode_after_release_gets_new_holdoff`: after a release, the next flip gets a fresh 60 s hold-off

`pytest`: 17 passed (14 existing + 3 new).

## Not covered

- Not yet tested on hardware. I'm commissioning a SolarFlow 2400 AC+ (zenSDK) with PV-surplus EV charging and will report back.
- The +4 W start walk after release (`device.py`, #1434) is untouched.

Refs #1525, #1461


igotsoul:
**Hardware test: SolarFlow 2400 AC+ (zenSDK), grid meter SMA Energy Meter via MQTT, patch applied at 12:07**

Load: induction hob cycling at ~85 s (≈10 s import of +550…+670 W, then off). Mode: `smart`.

| | 11:09–12:07 (1.4.4) | 12:07–12:16 (this PR) |
|---|---|---|
| Charge phases | **0** | **6** |
| Time share charging | 0 % | 25 % |
| Time share idle | 84 % | 75 % |
| Discharge phases | 30 (3–15 s each) | 0 |

Before the patch the battery did not charge once in 58 min, which is the unbounded re-arm described in #1525. With the patch it charges in every hob cycle.

Observed cycle with the patch: hob on (~10 s import) → hob off, ~−800 W surplus → **~60–70 s idle** (hold-off + ramp) → 15–20 s charging → hob on again.

Caveats:
- PV surplus increased after 12:07 (from ~−400 W to ~−1,400 W), so export in Wh is not comparable between the two windows. The charge/idle pattern is.
- Only 9 min with the patch so far. I'll add a longer window.

Remaining observation (not addressed by this PR): with a load period close to the hold-off, every cycle starts a new episode with a full 60 s, so the battery charges only ~25 % of the time. During charging, the start ramp (50 W + 4 W/tick) does not reach the available surplus within 15–20 s (P1 still ≈ −320 W on average while charging). Making `CHARGE_HOLDOFF` configurable might be worth considering.

cc @qpaz, this PR is based on your analysis in #1525. Thanks for the detailed write-up.
Also cc @Philippe26150 @uwegoexe-boop @krenk80 (#1525) and @mggevaer @CWILDMOUNTAIN @zoic21 (#1461), who reported similar behaviour.

PR #1665, opened 2026-09-21
Fix/charge holdoff debounce

### LINKED ISSUE #1664: [Bug] Manual mode: switching from smart to manual charge leaves the battery idle for ~1 minute
### Home Assistant Version

2026.9.3

v1.4.4

### Describe the bug

When I switch the Zendure Manager from `smart` to `manual` with `manual_power` already set to -2400 W, the battery does not start charging. It just stops: no charge, no discharge. If I then edit the manual power (e.g. -2399, then back to -2400) charging starts.

Related: #1525 (same hold-off in smart mode, closed as stale), #1663 and #1488 (open PRs that keep the 60 s delay on a single flip, so they do not cover this case).

### To Reproduce

1. Run the manager in `smart` mode for a while (the battery discharges at least once).
2. Set `number.zendure_manager_manual_power` to -2400.
3. Switch `select.zendure_manager_operation` to `manual`.
4. The battery stops. `operation_state` shows IDLE for about 60 s, then charging starts at 50 W and ramps up.

### Expected Behaviour

An explicit switch to `manual` should apply the requested power right away (normal 50 W kick-start aside). The anti-flapping hold-off has no purpose in manual mode, since the user fixes the direction. Resetting the hold-off timer on a mode change, or skipping it when the mode is `manual`, would fix it without touching smart mode.

SF2400 AC

serahug:
Fixes #1664.

`power_charge` gets a `debounce` parameter. `powerChanged` passes `debounce=False` for MANUAL (the user fixes the direction) and, slightly beyond the issue, for STORE_SOLAR, which never discharges and so has no flip to debounce. MATCHING and MATCHING_CHARGE keep the 60 s hold-off unchanged.

Two tests cover the hold-off with and without debounce.


igotsoul:
Nice, this is complementary to #1663: that PR bounds the hold-off in `smart` (MATCHING) mode, where the direction *is* measurement-driven and the hold-off currently re-arms on every flip without an upper bound (#1525). This PR removes it where there is no flip to debounce (MANUAL, STORE_SOLAR).

Both touch the same `# prevent hysteria` block in `power_charge`, so whichever gets merged second will need a small rebase. Happy to rebase #1663 on top of this one if it lands first. The two changes combine cleanly: `debounce` decides *whether* the hold-off applies, #1663 decides *how long* it can last.

Side note: this also helps setups that drive `manual` from an external EMS (e.g. evcc forcing a grid charge via an HA script), where a 60 s delay on every charge command is unexpected.

